PR #39, opened 2024-07-12
feat: pizzabot example with Step Functions

umaragu:
This sample contains the same example as in "complex-schema-tool-use"
The orchestration of invoking bedrock api, routing the tool and preparing the messages are part of Step Functions and Lambda.


schuettc:
Thanks for the contribution!

PR #158, opened 2025-03-18
Adding token efficient tool use beta example for Bedrock

jonathancaevans:
Add example of beta token efficient tool use on Bedrock


jawhnycooke:
Nice notebook!

PR #163, opened 2025-07-04
fix(claude_4_interleaved_thinking): rename additionalModelRequestFields to additional_request_fields

awsarron:
fix(claude_4_interleaved_thinking): rename additionalModelRequestFields to additional_request_fields


jawhnycooke:
fixed field based on updates to Strands

PR #198, opened 2026-06-30
Update converse stream example in sonnet 5 getting started notebook

najmia:
The `content` value is a plain string, but the Converse API requires it to be a list of content blocks.

**The problem:**
```python
"content": "List five best practices for IAM security."
```

**The fix:**
```python
"content": [{"text": "List five best practices for IAM security."}]
```

The Converse/ConverseStream API expects `messages[].content` to always be a list of `{"text": "..."}` objects — unlike the Messages API which accepts a bare string.


rodzanto (approved):
Looks good

PR #244, opened 2026-07-13
fix(claude-apps-gateway): region-agnostic deploy via global inference profiles + region-resolution & OIDC-secret fixes

bluedoors:
## What & why

Three fixes to the `claude-apps-gateway` **`deploy.sh` (CDK convenience path)**, surfaced while testing a deploy into a non-`us-east-1` region. They also apply to the `setup.sh` (CLI) path where relevant, so the two tracks stay in sync.

1. **Region resolution was inconsistent → deploy failed on a region mismatch.**
   `deploy.sh` pinned CDK calls + the ECR ARN to `BEDROCK_REGION` but left the bare `aws` CLI calls (`cloudformation`, `s3`, `sts`, `codebuild`) with no `--region`, so they followed the shell's `AWS_REGION`. A shell/`.env` mismatch built the stack in one region and looked for it in another. Now a single `DEPLOY_REGION` (stack: VPC/ALB/RDS/ECR/ECS/CodeBuild) is resolved `DEPLOY_REGION → BEDROCK_REGION → shell AWS_REGION → profile` and exported as `AWS_REGION`/`AWS_DEFAULT_REGION` so every `aws` call and CDK agree; `BEDROCK_REGION` now only drives the upstream `region:` + the inference-profile ARN.

2. **The OIDC client secret was never seeded on the `deploy.sh` path.**
   `.env`'s `OIDC_CLIENT_SECRET` was never written to Secrets Manager (the stack hardcoded `REPLACE_ME`), so OIDC login failed at the IdP with no obvious cause. `deploy.sh` now preflights it and seeds it into Secrets Manager after pass 2 (with a forced service redeploy so tasks pick it up), and the stack uses `generateSecretString` instead of a fixed placeholder so later deploys don't clobber the seeded value.

3. **The sample was implicitly US-only; now it deploys unchanged in any region.**
   The built-in model catalog resolves `us.anthropic.*` inference profiles, so a non-US `BEDROCK_REGION` deployed fine and then failed at inference with `400 The provided model identifier is invalid`. The config now uses an explicit `models:` block mapped to **global** cross-region inference profiles (`global.anthropic.*`, `auto_include_builtin_models: false`), and the IAM policy grants `inference-profile/global.anthropic.*`. Global profiles resolve from any commercial region, so the sample deploys unchanged regardless of `BEDROCK_REGION`. Network posture is unchanged — the gateway still talks only to the `BEDROCK_REGION` endpoint over the private VPC interface endpoint; cross-region routing is internal to Bedrock. The one tradeoff (global routes worldwide) is documented in `cdk/README` "Regions & data residency" with a geo-profile (`us.`/`eu.`/`apac.`) escape hatch.

   Default model set: Opus 4.8 / Sonnet 5 / Haiku 4.5. **Claude Fable 5 is opt-in** — on Bedrock it requires a non-default data-retention mode not enabled in every region (it returns `ValidationException: data retention mode 'default' is not available for this model` where the default three work), so shipping it by default would 400 in some regions. Documented how to add it back.

## Changes
- `cdk/scripts/deploy.sh`, `cdk/scripts/setup.sh` — region resolution / export; OIDC preflight + seed + redeploy; global-profile IAM ARN.
- `cdk/lib/claude-gateway-stack.ts`, `cdk/bin/app.ts` — `bedrockRegion` context; `global.anthropic.*` ARN; `generateSecretString` for the OIDC secret.
- `cdk/gateway.yaml.template` / `.example` — explicit `global.anthropic.*` `models:` block; Fable 5 commented opt-in.
- Docs — `cdk/README` (new "Regions & data residency"), repo `README`, `CLAUDE.md`, `.env.example`, `docs/{deployment,gotchas,upstream-watch,eks-notes}.md`.
- Tests — `cdk/test/gateway-stack.test.ts` (ARN → `global.`, cross-region ARN scoping, OIDC secret = generated placeholder / no static value).

## Validation
- **Unit/static:** `cd cdk && npm test` → 17/17; `./test/stamp-config.test.sh` 9/9; `./test/setup-helpers.test.sh` 14/14; `npx cdk synth`; `gateway.yaml.example` parses; `bash -n` + `shellcheck` clean.
- **Live (test account, both scripts, both regions):**
  - `setup.sh` → **us-east-1**: full deploy, ECS stable + ALB targets healthy, stamped `gateway.yaml` + task-role IAM both correctly `global.anthropic.*`, Fable absent.
  - `deploy.sh` (CDK) → **ap-southeast-2** (non-US): deployed **unchanged**, service stable + targets healthy, clean startup logs (no 403), task role scoped to `inference-profile/global.anthropic.*`, and `global.anthropic.claude-opus-4-8` invokes in-region.
  - (Interactive browser-SSO leg is environment-gated and validated separately.)

No linked issue — surfaced during live convenience-path testing.


scouturier:
Reviewed — region-resolution, OIDC-secret seeding, and global inference profile fixes all check out (confirmed `global.anthropic.*` profile IDs against AWS docs). One non-blocking note: `DEPLOY_REGION != BEDROCK_REGION` is documented as unsupported but not enforced in the stack (VPC Bedrock interface endpoint only lives in the deploy region) — worth a guard in a follow-up. Merging.

PR #245, opened 2026-07-14
Add ccwb-to-gateway migration skill

scouturier:
## What

Adds a project-scoped Claude Code skill, `ccwb-to-gateway`, under `claude-apps-gateway/.claude/skills/`. It generates concrete, AWS-focused migration plans for administrators moving from the **guidance-for-claude-code-with-amazon-bedrock** sample ("ccwb") to **claude-apps-gateway** (+ the optional **claude-apps-gateway-bootstrap** add-on).

## Why

ccwb is now in maintenance mode and claude-apps-gateway is the current recommendation, so admins running the credential-process/CloudFormation setup need a concrete path to the gateway's OIDC-SSO/proxy architecture. These are architecturally different systems (per-client AWS credentials vs. a server-held credential + bearer JWT), so the migration benefits from structured, deployment-specific guidance rather than generic advice.

## How it works

- **Interviews the admin** about their current deployment (IdP, auth mode, client mix, deployed stacks, network posture, rollout scope, account/region) before producing a full plan.
- **Reads all three repos live** rather than relying on memorized specifics — the embedded architecture-mapping table is explicitly framed as a starting map to verify against current source, with a "verify against" column, because both samples move quickly.
- **Scopes its output** — narrow questions get a focused answer; only cutover requests get the full nine-section plan.
- **Frames honestly** — ccwb as maintenance-mode with the gateway as the recommended path, without overselling the gateway sample (which self-describes as reference material, not a supported production artifact), and surfaces partial adoption as a valid outcome.

## Placement

Committed under `claude-apps-gateway/.claude/skills/` rather than the repo root because the migration is only relevant when working on the gateway sample; scoping it here means it auto-discovers for sessions in that directory.

## Notes

- Developed and iterated with the skill-creator eval loop; final pass was clean across the test cases (correct scoping, honest framing, source-grounded citations).
- Known follow-up (not in this PR): the skill currently assumes the three repos are cloned locally and asks for a path; a future enhancement could add a fetch/clone fallback for the external ccwb repo.


scouturier:
Superseded by #246, which is opened from a personal fork (scouturier/anthropic-on-aws) rather than a branch on the shared repo. Same commit, no changes.

PR #247, opened 2026-07-15
fix(claude-apps-gateway): deploy.sh stamps config from template so telemetry works

bluedoors:
## Problem

The recommended `deploy.sh` convenience path **silently disabled telemetry forwarding** by forking its own config.

- `deploy.sh` forked its own inline `gateway.yaml` (heredoc) instead of reusing `stamp-config.sh` + the tracked `cdk/Dockerfile`. That inline config had **no `telemetry:` block**, so the gateway never set `forward_to` and — per its own rule ("telemetry turns on only when `forward_to` AND `public_url` are both set") — never forwarded anything.
- The CDK task runs an **ADOT collector sidecar** (added in #243) ready to receive OTLP on `localhost:4318` and relay to CloudWatch, but with `forward_to` unset nothing ever reached it.

## Fix

**Align `deploy.sh` with `setup.sh`.** It now stamps `gateway.yaml` from the committed `gateway.yaml.template` via `stamp-config.sh` (which carries the `telemetry.forward_to: http://localhost:4318` sidecar block) and builds via the tracked distroless `cdk/Dockerfile`. Removes ~80 lines of duplicated config and closes the drift class where a template change could skip this path.

The sidecar and `CLAUDE_GATEWAY_ALLOW_LOOPBACK=1` (needed to push to the loopback receiver past the gateway's SSRF guard) both come from the CDK stack `deploy.sh` already deploys, so no extra task wiring is needed on this path.

**Refresh the docs** in `cdk/README.md` to the sidecar model: the resource table, the "How traffic flows" Telemetry leg, and the Cost table now describe one Fargate task running the gateway + ADOT sidecar — no separate service, no `:4318` ALB listener, no extra service cost.

## Verification (static)

- `bash -n` on `deploy.sh` + `setup.sh`
- `stamp-config.test.sh` (9) + `setup-helpers.test.sh` (14)
- `cdk` `npm test` (17) + `cdk synth` — sidecar image, `CLAUDE_GATEWAY_ALLOW_LOOPBACK`, and `cloudwatch:PutMetricData` all present in the synthesized template
- End-to-end stamp → valid YAML with `forward_to=http://localhost:4318` and `public_url` set, `${VAR}` secrets preserved, no `REPLACE_ME`/`@@` placeholders left

## Notes

- **Rebased onto `main` after #243** (standalone collector → sidecar). The README.md telemetry-section conflict was resolved in favor of #243's sidecar copy; this PR no longer edits the root README.
- The tracked Dockerfile pulls `debian:12-slim` + `gcr.io/distroless/...`; anonymous CodeBuild pulls can hit Docker Hub rate limits.
- Out of scope (found during earlier live testing, not fixed here): deploy.sh's CodeBuild IAM role scopes ECR push to the first run's `GATEWAY_NAME` and isn't re-scoped on a later name change.

### DIFF
--- a/claude-apps-gateway/cdk/README.md
+++ b/claude-apps-gateway/cdk/README.md
@@ -24,6 +24,7 @@ This CDK stack creates all the AWS infrastructure needed to run it:
 | **Security groups** | Network rules: ALB accepts HTTPS (443), ECS accepts traffic from ALB only (8080), RDS accepts traffic from ECS only (5432). |
 | **Route53 A record** | Points your gateway hostname at the ALB so developers can reach it by name. |
 | **CloudWatch log group** | Gateway logs go here. Boot messages, auth events, errors. |
+| **ADOT collector sidecar** | Runs *in the same Fargate task* as the gateway (not a separate service, no extra ALB listener). Receives per-user usage telemetry (OTLP) from the gateway on `localhost:4318` and forwards it to CloudWatch metrics (namespace `ClaudeGateway`) via SigV4 on the task role. The gateway container sets `CLAUDE_GATEWAY_ALLOW_LOOPBACK=1` so it can push to the loopback sidecar past its SSRF guard. See "Telemetry" below. |
 
 ## How traffic flows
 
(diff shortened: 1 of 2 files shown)


scouturier:
Reviewed — this is a clean de-dup fix, well aligned with the #243 sidecar migration. `deploy.sh` now stamps from the committed template via `stamp-config.sh` and builds the tracked `cdk/Dockerfile` (same path as `setup.sh`), so telemetry `forward_to` reaches the image and the drift class is closed. Verified the env contract matches `setup.sh`, the sidecar/loopback/PutMetricData all come from the CDK stack this deploys, and the stamp produces valid YAML end-to-end. Minor follow-ups (non-blocking): a durable Docker Hub rate-limit note for the CodeBuild path, and the ~$9 Fargate cost line is stale vs. 2×0.5 vCPU. Merging.

PR #248, opened 2026-07-16
fix(claude-apps-gateway): bootstrap CDK with imageReady=false to avoid pass-2 synth guard

qing-at-work:
## What

The gateway CDK docs tell first-time users to run `npx cdk bootstrap` with no context. But `cdk bootstrap` synthesizes the app before bootstrapping, and synth defaults to **pass 2** — which requires `publicUrl`/`zoneName`/`ingressCidr`/`certArn` and throws without them:

```
Error: Missing required context "publicUrl". For pass 2 deploy with: -c publicUrl=... ...
Or set imageReady=false for the pass-1 ECR-repo-only deploy.
```

## Fix

Add `-c imageReady=false` to the `cdk bootstrap` command in both places it appears, so synth short-circuits to the ECR-repo-only stack that needs none of that context:

- `claude-apps-gateway/cdk/README.md` (Option B — drive CDK yourself)
- `claude-apps-gateway/docs/deployment.md` (Track B — Pass 1)

Docs-only change. This matches the flag already used for the pass-1 `cdk deploy` that immediately follows.

### DIFF
--- a/claude-apps-gateway/cdk/README.md
+++ b/claude-apps-gateway/cdk/README.md
@@ -168,7 +168,7 @@ outputs; then pass 2 brings up the full stack (the service starts automatically
 at `desiredCount 2` — no manual scale-up):
 
 ```bash
-npx cdk bootstrap    # first time only
+npx cdk bootstrap -c imageReady=false    # first time only
 
 # Pass 1: ECR repo only
 npx cdk deploy -c imageReady=false \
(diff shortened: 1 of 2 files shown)


scouturier:
Verified — reproduced the exact `Missing required context "publicUrl"` error on bare `npx cdk synth` (pass 2 is the default per `bin/app.ts:66`), and confirmed `-c imageReady=false` short-circuits to the ECR-repo-only stack (`claude-gateway-stack.ts:87`) and synths cleanly. Both `cdk bootstrap` occurrences fixed, and those are the only two in the sample. Matches the flag already used on the pass-1 `cdk deploy` right below. Merging.

Issue #251, opened 2026-07-17
Runtime Error: Missing permissions with Gateway - ECS Task Execution Role


kayalvizhi-kandasamy:
**Situation**: Deployed the gateway service on Amazon ECS. TASK_ROLE gets created permissions for "_InvokeModel_" and "_InvokeModelWithResponseStream_". 

**Challenge**: After configuration, while prompting from Claude Code with any of the models the the gateway.yaml is configured with, the user is denied access.

**Additional details**:
CloudWatch logs for gateway mentions the below: 
`[gateway] 2026-07-17T09:54:40.434Z warn bedrock upstream 403 request_id=62100f49-e232-4630-a438-e95c128dd80e: 403 Model access is denied due to IAM user or service role is not authorized to perform the required AWS Marketplace actions (aws-marketplace:ViewSubscriptions, aws-marketplace:Subscribe) to enable access to this model. Refer to the Amazon Bedrock documentation for further details. Your AWS Marketplace subscription for this model cannot be completed at this time. If you recently fixed this issue, try again after 2 minutes.`

With the given message, if we add the below statement directly to the TASK_ROLE through IAM Console, the user is allowed from Claude Code CLI to invoke any of the models configured.
```
{
      "Sid": "VisualEditor0",
       "Effect": "Allow",
       "Action": [
                "aws-marketplace:ViewSubscriptions",
                "aws-marketplace:Subscribe"
        ],
        "Resource": "*"
}
```

**Recommendation**:
1. Adding the permissions through claude-apps-gateway/cdk/scripts/setup.sh (or) 
2. Mentioning it as a prerequisite in the doc would help

bluedoors:
Thanks for the detail here — and you're right that the docs could connect these dots better. Let me separate the two things: **why the error happens** (working as intended) and **how it should be fixed** (not by granting the task role Marketplace permissions).

[...]

PR #252, opened 2026-07-19
docs(claude-apps-gateway): clarify Bedrock model-access prereq + Marketplace 403

bluedoors:
## What

Makes the **Bedrock model-access prerequisite** discoverable and documents the AWS Marketplace `403` that a correct IAM setup still hits.

Addresses #251, where a customer deployed the gateway with a correct `bedrock:InvokeModel*` task-role policy and still got:

```
403 – Model access is denied due to IAM user or service role is not authorized to perform
the required AWS Marketplace actions (aws-marketplace:ViewSubscriptions, aws-marketplace:Subscribe)
```

## Why it's a docs fix, not a code fix

On Bedrock, Anthropic's models are AWS Marketplace offerings, so first use in an account requires a **one-time, account-level model-access grant (subscription)** — separate from IAM. This is a documented prerequisite (README prereq #3, `docs/deployment.md`, `docs/gotchas.md` §8), but none of those named the Marketplace `403` symptom, so it wasn't obvious that this error maps to the model-access step.

The tempting workaround — adding `aws-marketplace:Subscribe` to the ECS task role — makes the error disappear (the task self-subscribes on first call) but is a security anti-pattern:

- it grants a **one-time provisioning** action as a **standing** permission on a long-lived, network-exposed runtime identity;
- `Subscribe` is **account-wide** — a compromised container could subscribe the account to arbitrary paid Marketplace products and accept their EULAs, well outside an inference proxy's blast radius;
- procuring models is an **account-governance** decision, not a data-plane one.

That separation is exactly why model enablement is an out-of-band admin prereq the CDK deliberately does not automate. The right fix is the one-time console/API enablement; the task role then only needs `bedrock:InvokeModel*`.

## Changes

- **README prereq #3** — add a prominent **Model access** bullet: one-time account/admin action (not IAM, not a gateway responsibility), the `403`/Marketplace symptom, and an explicit "do not add Marketplace permissions to the task role" note.
- **`docs/gotchas.md` §8** — add the `aws-marketplace:*` `403` as an explicit symptom, the one-time admin fix, and a callout explaining why granting the task role `Subscribe` is the wrong posture (standing privilege, account-wide blast radius, separation of duties).

Docs-only; no behavior change.

## Testing

- `cd cdk && npm test` → 17/17 pass (unchanged).
- `python3 -c 'import yaml; yaml.safe_load(open("cdk/gateway.yaml.example"))'` → parses OK.

### DIFF
--- a/claude-apps-gateway/README.md
+++ b/claude-apps-gateway/README.md
@@ -84,6 +84,7 @@ You need an AWS account where they can create the following resources:
 - **Database**: An RDS PostgreSQL instance (db.t4g.micro is sufficient; the gateway stores only a few KB of sign-in state)
 - **Networking**: A VPC with private subnets, an internal ALB, and an imported ACM TLS certificate (use a public ACM cert to skip the first-login fingerprint prompt — see [`cdk/README.md`](cdk/README.md))
 - **IAM role**: The gateway's task role needs `bedrock:InvokeModel` and `bedrock:InvokeModelWithResponseStream` permissions on inference-profile and foundation-model ARNs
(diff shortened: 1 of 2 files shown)


scouturier (approved):
Approved: matches AWS's documented Bedrock Marketplace model-access flow; steers toward the supported admin-enablement path instead of a workaround. Docs-only, CI green.

PR #254, opened 2026-07-21
feat(claude-apps-gateway-bootstrap): resource-stripping OAuth proxy for Entra + AgentCore web search

chiararelandini:
## Problem

Claude Desktop / Cowork (and any spec-compliant MCP client) can't use an Amazon Bedrock **AgentCore Gateway** MCP server — e.g. the AgentCore web search tool — when the identity provider is **Microsoft Entra ID**.

On first connect the client discovers the gateway's Protected Resource Metadata (RFC 9728) and, per the MCP auth spec, adds an RFC 8707 `resource` parameter (the gateway URL) to the `/authorize` and `/token` requests. Entra does not implement RFC 8707 and rejects it:

```
AADSTS9010010: The resource parameter provided in the request doesn't match with the requested scopes.
```

There is no client-side configuration to suppress the parameter — Claude Desktop's managed-MCP `oauth` block, `mcp-remote`, and Claude Code all emit it (tracked in anthropics/claude-code#73460). With Amazon Cognito this "just works"; with Entra it is blocked.

## Solution

An **optional resource-stripping OAuth proxy** that fronts the gateway — the documented Microsoft/community workaround. Claude talks only to the proxy; the proxy mediates the OAuth handshake, **strips the `resource` parameter** before relaying to Entra, and reverse-proxies `/mcp` to the gateway. The gateway's `CUSTOM_JWT` authorizer is unchanged.

It deploys behind the **existing internal gateway ALB** (same pattern as the bootstrap add-on) via a path-prefix listener rule, and is **opt-in** (`-c enableWebsearchProxy=true`) so the default bootstrap deploy is unaffected.

## What's included

- `websearch-proxy/` — small Node/Express service + Dockerfile + README (problem, fix, architecture, endpoints, two-pass deploy)
- `cdk/lib/websearch-proxy-stack.ts` — `WebsearchProxyStack` (Fargate + target group + listener rule + SG rules), wired opt-in in `cdk/bin/app.ts`
- an example `managedMcpServers` entry in `bootstrap/config/bootstrap-config.example.json`

## How it was tested

Deployed to a test account (eu-south-1) behind the gateway's internal ALB and validated end-to-end through the real ALB:

- `/oauth/authorize` → 302 to Entra **without** `resource=`
- `/mcp` without a token → 401 carrying the proxy's own `resource_metadata`
- `/mcp` with a token → forwarded to the AgentCore gateway (gateway `x-amzn-requestid` returned)
- interactive Entra login on Claude Desktop → the web search MCP connects and tool calls succeed

## Notes

- The gateway's inbound authorizer is untouched (still validates the Entra token; `allowedAudience` stays the client id).
- For Claude Desktop the proxy only needs to strip `resource` (Desktop uses a static `clientId` and skips Dynamic Client Registration); the mock DCR / RFC 8414 endpoints are included for completeness / other clients.
- References (problem and fix): RFC 8707 / 9728 / 8414 / 7591, the Microsoft "Claude-ready Entra-protected MCP servers with APIM" guidance, and the "10 Entra + MCP incompatibilities" write-up — see the component README.

### DIFF
--- a/claude-apps-gateway-bootstrap/.gitignore
+++ b/claude-apps-gateway-bootstrap/.gitignore
@@ -1,7 +1,13 @@
 # Deployment-specific — copy the .example files
 .env
 bootstrap/config/bootstrap-config.json
+# Live top-level config edited per deployment (real hostnames/tenant/client ids);
+# the committed template is bootstrap/config/bootstrap-config.example.json.
+bootstrap-config.json
 cdk/cdk.context.json
+
+# Local, throwaway smoke-test scratch (not part of the deployable component)
+websearch-proxy/local-test/
 
 # Build artifacts & dependencies
 node_modules/
--- a/claude-apps-gateway-bootstrap/bootstrap/config/bootstrap-config.example.json
+++ b/claude-apps-gateway-bootstrap/bootstrap/config/bootstrap-config.example.json
@@ -37,6 +37,20 @@
       }
     },
     {
+      "name": "agentcore-websearch-entra-via-proxy",
+      "transport": "http",
+      "url": "https://claude-gateway.example.com/websearch/mcp",
+      "oauth": {
+        "clientId": "<entra-public-client-id>",
+        "authorizationServer": [
+          "https://claude-gateway.example.com/websearch"
+        ],
+        "scope": "openid offline_access api://<entra-public-client-id>/mcp.invoke",
+        "callbackHost": "localhost",
+        "callbackPort": 8990
+      }
+    },
+    {
       "name": "example-oauth-server",
       "transport": "http",
       "url": "https://<your-mcp-server>/mcp",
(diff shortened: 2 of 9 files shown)


scouturier:
Thanks for the thorough write-up and testing on this. Holding off on merging this, though: the root cause is that MCP clients (Claude Desktop, `mcp-remote`, Claude Code) unconditionally emit the RFC 8707 `resource` parameter, which Entra doesn't support — and that's already tracked upstream at the client level in anthropics/claude-code#73460.

A resource-stripping proxy is a reasonable workaround, but it's infrastructure we'd own and maintain indefinitely to paper over a client bug — and it would need to be torn out (or go stale) once the upstream fix lands. We'd rather wait for that fix than add a workaround with a shelf life.

Closing without merging for now — happy to revisit if the upstream issue stalls or if there's a stronger signal it won't be addressed client-side.

PR #257, opened 2026-07-23
fix(claude-apps-gateway): bound deploy.sh gateway probe; generalize VPN-only wording

bluedoors:
## What

Two small robustness/clarity fixes for the claude-apps-gateway worked example, found while running a live deploy.

### 1. Bound `deploy.sh`'s Step 5 verify curl

The final step curls `/.well-known/oauth-authorization-server` to confirm the gateway is live. The gateway sits behind an **internal** ALB (private IPs by design), and the curl had **no timeout** — so when `deploy.sh` runs from a host with no network path to the VPC, it hangs on the TCP connect, making an already-finished deploy look stuck.

- Add `--connect-timeout 5 --max-time 15` so it fails fast.
- Reword the guidance so a miss reads as **expected off-network** and does not imply the deploy failed (the stack/service are already up by that point).

### 2. Generalize "VPN-only" doc wording

A few spots in `cdk/README.md` framed VPN as the *only* way to reach the internal ALB. Reworded to "private network path to the VPC (VPN / Direct Connect / Transit Gateway)", matching the convention already used in `cdk/scripts/setup.sh` and `docs/deployment.md`. Legitimately VPN-specific content (the Client VPN appendix in `docs/connectivity.md`, teardown steps, `INGRESS_CIDR` client-CIDR notes) was left unchanged.

## Verification

- `bash -n cdk/scripts/deploy.sh` clean; `shellcheck` clean (only the pre-existing SC1091 info on the sourced `.env`).
- Docs-only for the wording change.

### DIFF
--- a/claude-apps-gateway/cdk/README.md
+++ b/claude-apps-gateway/cdk/README.md
@@ -353,9 +353,9 @@ easy teardown is data durability. Before using it with real workloads, edit
   internal ALB (its hostname must resolve to private IPs only, which is why the ALB is
   internal in the first place).
 
-## Testing without a VPN (local workaround)
+## Testing without VPC network access (local workaround)
 
-In production, the gateway runs behind an internal ALB and developers access it through VPN. But for testing on your own laptop without VPN access to the VPC, you can run the gateway locally.
+In production, the gateway runs behind an internal ALB and developers reach it over a private network path to the VPC (VPN, Direct Connect, Transit Gateway, etc.). But for testing on your own laptop without that access, you can run the gateway locally.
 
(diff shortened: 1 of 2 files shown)


scouturier (approved):
Approved: probe timeout fix is correctly bounded — Step 5 runs after ecs wait services-stable already confirmed the deploy, so no false-negative risk. Reworded messaging correctly separates 'still starting' from 'unreachable network'. VPN wording generalization aligns with terminology already used elsewhere in the repo, no loss of specificity. No security concerns, diff scoped to the 2 claimed files, CI green, no conflict with merged predecessors.

PR #264, opened 2026-07-26
feat(claude-apps-gateway): offer Claude Opus 5 in the model catalog

bluedoors:
## What

Adds **Claude Opus 5** (`claude-opus-5`) to the example's model catalog, replacing Claude Opus 4.8. Opus 5 is now the recommended model for agentic coding / enterprise work and is available on Amazon Bedrock; Opus 4.8 has moved to Anthropic's legacy tier.

## Changes

- **`cdk/gateway.yaml.template` + `cdk/gateway.yaml.example`** — replace `claude-opus-4-8` with `claude-opus-5` in both the `managed.policies.availableModels` picker allowlist and the `models:` catalog. The catalog entry maps to `global.anthropic.claude-opus-5`, matching the dateless global-inference-profile pattern the `claude-sonnet-5` entry already uses.
- **Docs** (`cdk/README.md`, `README.md`, `docs/deployment.md`) — update the "shipped catalog" statement and the catalog / short-alias examples that mirror it.

## Why no IAM or version change

- **No IAM/CDK change:** the task role already grants `bedrock:InvokeModel*` on the wildcards `inference-profile/global.anthropic.*` and `foundation-model/anthropic.*`, which cover Opus 5. The CDK tests assert those wildcards (not specific model IDs), so nothing breaks.
- **No pinned-CLI bump:** the gateway resolves operator-defined model IDs straight to Bedrock inference profiles (`auto_include_builtin_models: false`), so a new model id is neither a new gateway subcommand nor a new managed-settings key. `CLAUDE_VERSION` stays at 2.1.218.
- **Operator prerequisite unchanged:** enable Bedrock model access for the model you list (already documented). Confirm the exact global profile id with `aws bedrock list-inference-profiles` per the template's existing caveat.

Deliberately left as-is: illustrative policy/failover snippets that already show non-shipped models (e.g. `sonnet-4-6`, on-demand profiles), workshop teaching material, and the `LESSONS.md` live-deploy log (historical record).

## Verification

`gateway.yaml.example` parses; the `stamp-config` suite stamps the edited template to valid YAML (9); `setup-helpers` (14); `cdk synth` + jest (17) all pass. No live AWS account exercised (per repo convention).

### DIFF
--- a/claude-apps-gateway/README.md
+++ b/claude-apps-gateway/README.md
@@ -240,9 +240,9 @@ upstreams:
 # region-agnostic. (For data residency, swap global. for a geo prefix: us./eu./apac.)
 auto_include_builtin_models: false
 models:
-  - id: claude-opus-4-8
-    label: Claude Opus 4.8
-    upstream_model: { bedrock: global.anthropic.claude-opus-4-8 }
+  - id: claude-opus-5
+    label: Claude Opus 5
+    upstream_model: { bedrock: global.anthropic.claude-opus-5 }
   - id: claude-haiku-4-5
     label: Claude Haiku 4.5   # no short alias — dated profile id
     upstream_model: { bedrock: global.anthropic.claude-haiku-4-5-20251001-v1:0 }
--- a/claude-apps-gateway/cdk/README.md
+++ b/claude-apps-gateway/cdk/README.md
@@ -312,7 +312,7 @@ inference confined to a specific geography, switch the catalog off global:
 - A geo (or in-region) profile only routes within that geography, so pick a
   `BEDROCK_REGION` the profile actually spans.
 
-**Default model set.** The shipped catalog is Claude Opus 4.8, Sonnet 5, and Haiku 4.5
+**Default model set.** The shipped catalog is Claude Opus 5, Sonnet 5, and Haiku 4.5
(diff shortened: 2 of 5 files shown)


bluedoors:
Superseded by #282, which lands this same Opus 5 catalog swap as part of the 2.1.229 pin bump.

Closing rather than rebasing because #282 is a strict superset: it makes the identical
`claude-opus-4-8` → `claude-opus-5` change in `availableModels` and the `models:` catalog,
across the same 5 files, and adds the pin bump plus the `chatTabEnabled` fix that landed in
2.1.227.

Two things #282 settles that this PR left open:

- **`global.anthropic.claude-opus-5` is confirmed ACTIVE on Bedrock**, with a successful
  `converse` call for it and for every other model in the shipped catalog. This PR reasoned
  from the wildcard IAM grant and the dateless-profile naming pattern, which was sound but
  unverified — a model being current in Claude Code is a separate claim from a Bedrock
  inference profile existing under the `global.` prefix. `docs/upstream-watch.md` now says to
  check both.
- **The premise that no CLI bump was needed** held for the model id alone, but the pin was
  11 releases behind by then, so shipping Opus 5 on 2.1.218 would have left the
  `chatTabEnabled` dead end (#83723 / PR #277) in place. #282 does both together.

No changes lost.

PR #266, opened 2026-07-27
fix(claude-apps-gateway): make VPC reuse and repeat deploys safe

bluedoors:
## Scope: deliberately minimal

These changes are kept as small as possible so this stays a **worked example, not a production artifact**. Every addition is either a fix for something that loses data / deletes infrastructure, or a fix for VPC reuse not actually working as documented. Where a scenario genuinely isn't supported — multi-region concurrency, locked endpoint security groups, renaming a live deployment, proving ownership of resources in a shared VPC — this follows the repo's existing convention and **says so in the docs rather than automating around it**.

That's a deliberate line. An earlier draft of this branch grew an ownership-tag scheme and a wider status-handling matrix; both were removed in favour of documenting the constraint.

Net: **+556 / −57** across 8 files, one new file (a test), no dependency changes.

### 1. Re-running `deploy.sh` destroyed the data plane

Pass 1 (`-c imageReady=false`) returns immediately after creating the ECR repo, so its template holds **one** resource against pass 2's **53**. `deploy.sh` ran it unconditionally at Step 1 of *every* invocation — so a second run told CloudFormation to delete everything absent from the repo-only template: the RDS instance (`deletionProtection: false`, `RemovalPolicy.DESTROY`, 1-day backups, so the gateway's store went with it), the ALB, the ECS service, all three secrets, the log group, and the VPC.

Fix: gate Step 1 on stack status, with the **destructive** branch holding the narrow allowlist so unknown and future statuses fail closed — run pass 1 only when the stack is absent or provably empty, skip to pass 2 otherwise. The status query is itself a safety control, so it doesn't fail open either: only an explicit `ValidationError`/"does not exist" counts as absent, and a throttle, expired credential, or network blip aborts rather than guessing. stdout and stderr are captured separately, so a CLI that succeeds while printing a warning to stderr can't prepend that text to the status and send a recoverable `ROLLBACK_COMPLETE` stack down the unknown-status path.

The two decisions are pure string→branch logic, so they're behind a `DEPLOY_SH_LIB_ONLY` lib gate (mirroring `setup.sh`'s `SETUP_SH_LIB_ONLY`) and asserted in `test/deploy-helpers.test.sh` — 31 assertions, no AWS account.

Everything *other* than the pass-1 decision is left to CDK on purpose, rather than re-implementing CloudFormation's status list. The comment at the gate records what that actually costs, read off the pinned CDK (2.1129) rather than assumed: it waits out `*_IN_PROGRESS`, deletes and recreates a failed first create (`ROLLBACK_COMPLETE` / `ROLLBACK_FAILED`), and **asks** before rolling a fail-paused stack forward (`CREATE_FAILED` / `UPDATE_FAILED` / `UPDATE_ROLLBACK_FAILED`). `--require-approval never` does not waive that prompt — only `--force` does — so an interactive run stops for a y/n and a non-TTY run fails with `TtyNotAttached`, in both cases after the image build.

The hand-driven CDK track can't be gated in code, so it gets a short `[!CAUTION]` at the pass-1 command.

### 2. Teardown could destroy a shared VPC's resources

`docs/teardown.md`'s network block deleted every endpoint, subnet, non-default SG, route table, NAT, and IGW filtered **only by vpc-id**, guarding just the final `delete-vpc`. Every loop is now scoped to the `Project=$P` tag `setup.sh` already stamps, and `setup.sh` now tags the VPC endpoints it creates (previously the one untagged resource).

Tag-scoping alone isn't ownership, though: `setup.sh` *adopts* a matching subnet, IGW, NAT, or SG when it finds one and tags only what it **creates**, so an adopted resource keeps whatever tags you gave it. Rather than invent an ownership marker, the automated block is now explicitly for a `setup.sh`-created VPC, and reused VPCs get a short "Reused VPC" section: five `describe` commands listing candidates with their `Name`/`Project` tags, plus the names `setup.sh` uses, so you delete what you recognise by hand.

`delete-vpc` is gated twice regardless — it skips when you supplied `VPC_ID` (recorded before the lookup can clobber it) and skips anything not tagged `Project=$P` — so a stray paste can't take the VPC. That reuse flag is decided **once**, on the first paste, because the lookup writes to `VPC_ID` too: re-deriving it would make a re-paste of the block (which the SG-straggler note asks for) read the looked-up id as "you supplied it" and skip `delete-vpc` on a VPC `setup.sh` created.

### 3. VPC reuse didn't work as documented

- **Fixed subnet CIDRs.** `setup.sh` creates subnets at `10.20.0.0/24`, `10.20.1.0/24`, `10.20.10.0/24`, `10.20.11.0/24`, so a reused VPC must be `10.20.0.0/16` or a superset with those free. The CDK track imports existing subnets, so any CIDR works. This wasn't stated anywhere.
- **Endpoint security groups.** With `createVpcEndpoints=false`, neither track touches the reused **interface** endpoints' SGs — you must allow 443 from the task SG yourself, or tasks time out fetching secrets, images, and logs. S3 is a *gateway* endpoint: route-table association, no SG, no 443. The docs previously implied otherwise. The flag is CDK-track-only, which the docs now say: `setup.sh` describes before creating, so it adopts endpoints already in the VPC and needs no equivalent.
- **Ordering.** CDK's `TaskSg` exists only in pass 2, and a pass 2 that can't stabilize **rolls back and deletes the very SG you were about to authorize** — so the documented deploy → authorize → redeploy sequence was impossible. Pass 2 now accepts rollback being disabled (`NO_ROLLBACK=1` for `deploy.sh`, `--no-rollback` by hand), off by default because the resulting `UPDATE_FAILED` must be retried or abandoned manually. Set it per run: an update requiring resource *replacement* can't be deployed with rollback disabled, so a leftover export makes an unrelated later change fail confusingly.

### 4. `deploy.sh` CodeBuild bucket, source, and IAM

- **Region-suffix the S3 build bucket** (`…-${ACCOUNT_ID}-${DEPLOY_REGION}`): CodeBuild requires its source bucket in the project's region, so an account-global name broke any second-region build. `docs/teardown.md` now has a CodeBuild-artifacts section covering the role, the project, the region-suffixed bucket, and the unsuffixed one an upgrade strands.
- **Assert the project's source on every run** (`create-project` first time, else `update-project`). A project from the previous script still pointed at the old unsuffixed bucket, and since the role now grants only the new one, `start-build` failed downloading its source on the first run after upgrading — i.e. for every existing deployment.
- **Re-assert the role + policy every run** and move the IAM-propagation sleep after every policy assertion, so a repo/region change can't hit the build before IAM propagates (intermittent ECR `AccessDenied`). Both calls now report their own failures: `create-role`'s blanket `|| true` is narrowed to `EntityAlreadyExists`, and `put-role-policy` no longer discards stderr — silencing it on a call that runs every deploy turned a real IAM failure into the script dying after "Step 3/5" with nothing printed.
- **Refuse a `GATEWAY_NAME` change** against an existing stack: pass 2 would replace the ECR repo and the service would pull from an empty one. It can't be fixed by pushing first, since the only template that creates the renamed repo ahead of the push is the destructive pass 1. A clear refusal beats a confusing ECS pull error — and because pass 1 can now be skipped, a stack with no `EcrRepositoryUri` output is caught explicitly rather than reported as that rename ( `--output text` yields `None` for a null query and an empty string for an unmatched filter; both are guarded).

### 5. Docs reconciled

The reuse guidance had ended up restated in four places, with reference-table cells reaching 111 words in an earlier draft of this branch. There's now one "Reusing an existing VPC" section in `docs/deployment.md`; both tracks and both `cdk/README.md` rows point at it, and each table cell is one sentence plus that link. Measured against `main` the four cells are *longer*, not shorter — they gained the cross-reference (`createVpcEndpoints` 40 → 54 words, `CREATE_VPC_ENDPOINTS` 21 → 40, `DEPLOY_REGION` 28 → 46, `GATEWAY_NAME` 27 → 47); the consolidation is against the draft, not against `main`.

Also reconciled while in these files: the `NOTE` claiming `deploy.sh` uses "its own inline Dockerfile and config" and "does not pass CDK context" — neither is true (it shares the tracked `Dockerfile` + `stamp-config.sh` and builds the full `CDK_CTX` from `.env`). Three more copies of that stale claim in `cdk/README.md` are fixed the same way; the real remaining deltas are the unverified staged binary and the `:latest` tag. Two stale anchors fixed (`#regions`, and `cdk/README.md`'s binary-staging link).

## Verification

No AWS account required; all local/static per `CLAUDE.md`:

- `bash -n` + `shellcheck` on `setup.sh` and `deploy.sh` (only the pre-existing `SC1091` on `source .env`)
- `./test/deploy-helpers.test.sh` — **31 passed**, new in this PR: the pass-1 allowlist (absent / `ROLLBACK_COMPLETE` / `REVIEW_IN_PROGRESS` / `DELETE_COMPLETE` → run; `CREATE_COMPLETE` / `UPDATE_COMPLETE` / `UPDATE_FAILED` / `IMPORT_COMPLETE` / `*CLEANUP_IN_PROGRESS` / `DELETE_FAILED` / `CREATE_FAILED` / an invented future status → skip), the fail-closed status query (`ExpiredToken` / `Throttling` / `AccessDenied` / unreachable endpoint / empty stderr → abort; only `ValidationError`/"does not exist" → first deploy), and that a stderr warning alongside a successful query leaves the status recoverable
- `./test/stamp-config.test.sh` (**9 passed**), `./test/setup-helpers.test.sh` (**19 passed**)
- `cd cdk && npm test` (**17 passed**), `npx cdk synth` for both passes
- All three shell suites also run under macOS's system `/bin/bash` (3.2.57)
- The three `delete-vpc` branches run against a stubbed `aws`: dedicated-and-tagged deletes, reused-untagged skips, reused-**and**-tagged skips
- The five new teardown listing queries compiled with `jmespath` against a sample response
- Every cross-reference in the touched docs resolved against GitHub's heading-slug rules
- `gateway.yaml.example` parses as YAML

The teardown block is documentation and this repo doesn't test markdown, so the stubbed-`aws` harness for it stayed throwaway; the pass-1 gate's matrix did not — it's the committed suite above.

End-to-end smoke testing against a real account + OIDC IdP remains the operator's, per the gateway quickstart.

### DIFF
--- a/claude-apps-gateway/cdk/scripts/deploy.sh
+++ b/claude-apps-gateway/cdk/scripts/deploy.sh
@@ -95,9 +95,59 @@ fi
 # The ECS service can't start until its image exists, so the stack splits the
 # deploy in two: pass 1 creates just the ECR repo; we build+push; pass 2 (below)
 # brings up the full stack. See cdk/README.md "CDK context variables".
-echo "Step 1/5: Pass 1 — creating the ECR repository (CDK)..."
-npx cdk deploy --require-approval never -c imageReady=false "${CDK_CTX[@]}"
-echo "✅ ECR repository created"
+#
+# Pass 1 is FIRST-DEPLOY ONLY. Its template holds exactly one resource (the ECR
+# repo), so deploying it over a stack that already reached pass 2 makes
+# CloudFormation delete the ~50 resources absent from it: the RDS instance
+# (deletionProtection false + RemovalPolicy.DESTROY + 1-day backups, so the
+# gateway's store is gone), the ALB, the ECS service, the secrets, the log group,
+# the VPC. Re-running deploy.sh must therefore go straight to build + pass 2,
+# which is also what preserves TaskSg and any 443 rules you authorized on reused
+# interface endpoints (see NO_ROLLBACK below).
+#
+# The status query itself is a safety control, so it must not fail open: a
+# throttle, expired credential, or network blip would yield an empty status, look
+# like "no stack yet", and run the destructive pass against a live one. Treat ONLY
+# an explicit not-found as absent, and abort on anything else.
+set +e
+STACK_QUERY=$(aws cloudformation describe-stacks --stack-name ClaudeGatewayStack \
+  --query 'Stacks[0].StackStatus' --output text 2>&1)
+STACK_RC=$?
+set -e
+if [ "$STACK_RC" -eq 0 ]; then
+  STACK_STATUS="$STACK_QUERY"
+else
+  case "$STACK_QUERY" in
+    *ValidationError*"does not exist"*)
+      STACK_STATUS=""   # genuinely absent → first deploy
+      ;;
+    *)
+      echo "❌ Could not determine the ClaudeGatewayStack status, so deploy.sh cannot"
+      echo "   tell a first deploy from an existing one — refusing to continue, because"
+      echo "   guessing wrong runs a repo-only deploy that would delete the RDS instance"
+      echo "   and the rest of the running stack. Fix the AWS call and re-run:"
+      echo "   $STACK_QUERY"
+      exit 1
+      ;;
+  esac
+fi
+# Running pass 1 is the destructive branch, so IT gets the narrow allowlist: only a
+# stack that is absent or provably holds nothing. Every other status skips to pass 2,
+# which is always the safe default — if the stack is mid-operation or wedged, `cdk
+# deploy` says so itself instead of us enumerating CloudFormation's status list.
+case "$STACK_STATUS" in
+  # No stack, or one holding no resources. ROLLBACK_COMPLETE is a failed FIRST create
+  # (nothing provisioned); CDK deletes and recreates it. REVIEW_IN_PROGRESS is a
+  # change-set-only stack.
+  ""|ROLLBACK_COMPLETE|REVIEW_IN_PROGRESS|DELETE_COMPLETE)
+    echo "Step 1/5: Pass 1 — creating the ECR repository (CDK)..."
+    npx cdk deploy --require-approval never -c imageReady=false "${CDK_CTX[@]}"
+    echo "✅ ECR repository created"
+    ;;
+  *)
+    echo "Step 1/5: stack exists ($STACK_STATUS) — skipping pass 1 (a repo-only deploy would delete the ALB/RDS/service)."
+    ;;
+esac
 echo ""
 
 # --- Step 2: Get outputs ---
@@ -109,35 +159,72 @@ ECR_URI=$(aws cloudformation describe-stacks --stack-name ClaudeGatewayStack \
 # alone, so a name mismatch with the stack can't break the push).
 REPO_NAME="${ECR_URI##*/}"
 echo "   ECR: $ECR_URI (repo: $REPO_NAME)"
+
+# Renaming a deployed gateway in place isn't supported: pass 2 would replace the ECR
+# repo, leaving the service pulling from an empty one. Fail here rather than at a
+# confusing ECS pull error. See cdk/README.md (GATEWAY_NAME).
+if [ "$REPO_NAME" != "$GATEWAY_NAME" ]; then
+  echo "❌ GATEWAY_NAME is '$GATEWAY_NAME' but this stack's repo is '$REPO_NAME'. Set it"
+  echo "   back, or tear down (docs/teardown.md) and deploy fresh under the new name."
+  exit 1
+fi
 echo ""
 
 # --- Step 3: Build and push image ---
 echo "Step 3/5: Building and pushing gateway image..."
 
 ACCOUNT_ID=$(aws sts get-caller-identity --query Account --output text)
-BUCKET="claude-gateway-build-${ACCOUNT_ID}"
+# Region-suffixed: CodeBuild requires its S3 source bucket in the PROJECT's region,
+# so the bucket name must vary by region (an account-global name would leave the
+# bucket in the first region and fail a build in any other). See the single-region
+# scope note below.
+BUCKET="claude-gateway-build-${ACCOUNT_ID}-${DEPLOY_REGION}"
 # Create the S3 build bucket BEFORE the CodeBuild project, which references it as
 # its source location — create-project fails with "Bucket ... does not exist"
 # otherwise (only bites a first-ever run, where the project doesn't yet exist).
-aws s3 mb "s3://$BUCKET" 2>/dev/null || true
+aws s3 mb "s3://$BUCKET" --region "$DEPLOY_REGION" 2>/dev/null || true
 
-# Check if CodeBuild project exists, create if not
+# CodeBuild role + policy. The project has a FIXED name (CodeBuild projects are
+# region-scoped, so the same name in another region is a separate project); the
+# role has a FIXED name and IAM is global. The inline policy is scoped to THIS
+# deploy's ECR repo + region, so re-assert it on EVERY run (idempotent overwrite)
+# instead of only at first project creation — otherwise redeploying with a
+# different gatewayName leaves the role authorized for the previous repo and the
+# build fails with an opaque ECR AccessDenied.
+#
+# SCOPE: deploy.sh targets ONE active region per account. The shared global role
+# means concurrent deployments in multiple regions (or under different gateway
+# names) overwrite each other's policy — give the role + project per-region/-name
+# suffixes if you truly need them side by side. See cdk/README.md (DEPLOY_REGION).
+aws iam create-role --role-name claude-gateway-codebuild \
+  --assume-role-policy-document '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"Service":"codebuild.amazonaws.com"},"Action":"sts:AssumeRole"}]}' 2>/dev/null || true
+
+aws iam put-role-policy --role-name claude-gateway-codebuild --policy-name build-perms \
+  --policy-document "{\"Version\":\"2012-10-17\",\"Statement\":[{\"Effect\":\"Allow\",\"Action\":[\"s3:GetObject\",\"s3:ListBucket\"],\"Resource\":[\"arn:aws:s3:::${BUCKET}\",\"arn:aws:s3:::${BUCKET}/*\"]},{\"Effect\":\"Allow\",\"Action\":[\"ecr:GetAuthorizationToken\"],\"Resource\":\"*\"},{\"Effect\":\"Allow\",\"Action\":[\"ecr:BatchCheckLayerAvailability\",\"ecr:GetDownloadUrlForLayer\",\"ecr:BatchGetImage\",\"ecr:PutImage\",\"ecr:InitiateLayerUpload\",\"ecr:UploadLayerPart\",\"ecr:CompleteLayerUpload\"],\"Resource\":\"arn:aws:ecr:${DEPLOY_REGION}:${ACCOUNT_ID}:repository/${REPO_NAME}\"},{ \"Effect\":\"Allow\",\"Action\":[\"logs:CreateLogGroup\",\"logs:CreateLogStream\",\"logs:PutLogEvents\"],\"Resource\":\"*\"}]}" 2>/dev/null
+
+# IAM is eventually consistent, so wait AFTER (re)asserting the policy — every run,
+# not only at project creation. A prior fix ran this sleep only when the project
+# was first created, so a policy change on an existing project (new repo/region)
+# could hit the build before it propagated → intermittent ECR AccessDenied.
+sleep 10  # IAM propagation for the (re)asserted role policy
+
+# The project's S3 source must track $BUCKET, so assert it on EVERY run — create it
+# on a first deploy, else UPDATE an existing project's source. A project created by
+# an earlier version of this script points at the old unsuffixed
+# claude-gateway-build-${ACCOUNT_ID} bucket; since we now upload to the
+# region-suffixed bucket and the role policy above grants only that one, leaving the
+# stale source in place makes start-build fail while downloading its source.
+CB_SOURCE="{\"type\":\"S3\",\"location\":\"${BUCKET}/\",\"buildspec\":\"buildspec.yml\"}"
 if ! aws codebuild batch-get-projects --names claude-gateway-build --query "projects[0].name" --output text 2>/dev/null | grep -q claude-gateway-build; then
   echo "   Creating CodeBuild project..."
-
-  # Create IAM role for CodeBuild
-  aws iam create-role --role-name claude-gateway-codebuild \
-    --assume-role-policy-document '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"Service":"codebuild.amazonaws.com"},"Action":"sts:AssumeRole"}]}' 2>/dev/null || true
-
-  aws iam put-role-policy --role-name claude-gateway-codebuild --policy-name build-perms \
-    --policy-document "{\"Version\":\"2012-10-17\",\"Statement\":[{\"Effect\":\"Allow\",\"Action\":[\"s3:GetObject\",\"s3:ListBucket\"],\"Resource\":[\"arn:aws:s3:::claude-gateway-build-${ACCOUNT_ID}\",\"arn:aws:s3:::claude-gateway-build-${ACCOUNT_ID}/*\"]},{\"Effect\":\"Allow\",\"Action\":[\"ecr:GetAuthorizationToken\"],\"Resource\":\"*\"},{\"Effect\":\"Allow\",\"Action\":[\"ecr:BatchCheckLayerAvailability\",\"ecr:GetDownloadUrlForLayer\",\"ecr:BatchGetImage\",\"ecr:PutImage\",\"ecr:InitiateLayerUpload\",\"ecr:UploadLayerPart\",\"ecr:CompleteLayerUpload\"],\"Resource\":\"arn:aws:ecr:${DEPLOY_REGION}:${ACCOUNT_ID}:repository/${REPO_NAME}\"},{ \"Effect\":\"Allow\",\"Action\":[\"logs:CreateLogGroup\",\"logs:CreateLogStream\",\"logs:PutLogEvents\"],\"Resource\":\"*\"}]}" 2>/dev/null
-
-  sleep 10  # Wait for IAM propagation
-
   aws codebuild create-project --name claude-gateway-build \
-    --source "{\"type\":\"S3\",\"location\":\"${BUCKET}/\",\"buildspec\":\"buildspec.yml\"}" \
+    --source "$CB_SOURCE" \
     --artifacts '{"type":"NO_ARTIFACTS"}' \
     --environment '{"type":"LINUX_CONTAINER","image":"aws/codebuild/standard:7.0","computeType":"BUILD_GENERAL1_SMALL","privilegedMode":true}' \
+    --service-role "arn:aws:iam::${ACCOUNT_ID}:role/claude-gateway-codebuild" >/dev/null
+else
+  aws codebuild update-project --name claude-gateway-build \
+    --source "$CB_SOURCE" \
     --service-role "arn:aws:iam::${ACCOUNT_ID}:role/claude-gateway-codebuild" >/dev/null
 fi
 
@@ -226,8 +313,19 @@ echo ""
 # The image now exists in ECR, so pass 2 brings up the service (desiredCount 2)
 # behind the internal ALB. cdk deploy blocks until the service is stable, so no
 # manual `update-service` scale-up is needed.
+#
+# NO_ROLLBACK=1 adds --no-rollback (CloudFormation DisableRollback). Needed when
+# reusing a VPC whose interface endpoints don't yet allow 443 from the task SG: the
+# service can't stabilize, and the default rollback deletes the just-created TaskSg
+# — the very SG you have to authorize on those endpoints. Retaining it makes the
+# documented deploy → authorize → redeploy sequence possible. Off by default: the
+# failure leaves the stack in UPDATE_FAILED, which you then have to either retry or
+# abandon by hand (`npx cdk rollback` / `aws cloudformation rollback-stack` —
+# continue-update-rollback does NOT apply, it only takes UPDATE_ROLLBACK_FAILED).
+CDK_DEPLOY_FLAGS=(--require-approval never)
+[ "${NO_ROLLBACK:-0}" = "1" ] && CDK_DEPLOY_FLAGS+=(--no-rollback)
 echo "Step 4/5: Pass 2 — deploying the full stack (CDK)..."
-npx cdk deploy --require-approval never -c imageReady=true "${CDK_CTX[@]}"
+npx cdk deploy "${CDK_DEPLOY_FLAGS[@]}" -c imageReady=true "${CDK_CTX[@]}"
 echo "✅ Full stack deployed"
 echo ""
 
--- a/claude-apps-gateway/cdk/scripts/deploy.sh
+++ b/claude-apps-gateway/cdk/scripts/deploy.sh
@@ -18,6 +18,61 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 PROJECT_DIR="$(dirname "$SCRIPT_DIR")"
 cd "$PROJECT_DIR"
 
+# ──────────────────────────────────────────────────────────────────────────────
+# Helpers for the pass-1 safety gate (Step 1). Defined up here, before any config
+# or AWS call, so tests can source them in isolation — see the DEPLOY_SH_LIB_ONLY
+# gate below and test/deploy-helpers.test.sh. The gate decides whether to run a
+# deploy that DELETES a live data plane, so its string→branch logic is tested
+# rather than reasoned about.
+# ──────────────────────────────────────────────────────────────────────────────
+
+# classify_stack_query — turn a `describe-stacks` result into a stack status.
+# Prints the status on success, prints NOTHING when the stack is genuinely absent,
+# and returns non-zero when the query failed for any OTHER reason — callers must
+# abort on that. This query is itself a safety control, so it must not fail open: a
+# throttle, an expired credential, or a network blip would otherwise look like "no
+# stack yet" and run the destructive pass against a live stack.
+#
+# stdout and stderr are passed in SEPARATELY on purpose. The CLI can print a
+# warning to stderr while still succeeding (urllib3's NotOpenSSLWarning on some
+# macOS Pythons); folding stderr into stdout would prepend that text to the status,
+# and a recoverable ROLLBACK_COMPLETE stack would then read as an unknown status and
+# skip the pass 1 it needs.
+classify_stack_query() {
+  local rc="$1" out="$2" err="$3"
+  if [ "$rc" -eq 0 ]; then
+    printf '%s' "$out"
+    return 0
+  fi
+  case "$err" in
+    *ValidationError*"does not exist"*) return 0 ;;  # genuinely absent → first deploy
+    *) return 1 ;;
+  esac
+}
+
+# needs_pass_one — should the ECR-repo-only pass 1 run for this stack status?
+# Running pass 1 is the DESTRUCTIVE branch, so it gets the narrow allowlist: only a
+# stack that is absent or provably holds nothing. Every other status — including one
+# neither this script nor today's CloudFormation knows about — skips to pass 2,
+# which is always the safe default.
+needs_pass_one() {
+  case "$1" in
+    # "" is no stack at all. ROLLBACK_COMPLETE is a failed FIRST create (nothing
+    # provisioned); CDK deletes and recreates it. REVIEW_IN_PROGRESS is a
+    # change-set-only stack. DELETE_COMPLETE is listed because it belongs in the
+    # "holds nothing" set, but nothing relies on it: describe-stacks by NAME never
+    # returns a deleted stack (that needs the stack ID), so the arm is unreachable
+    # from the call site below.
+    ""|ROLLBACK_COMPLETE|REVIEW_IN_PROGRESS|DELETE_COMPLETE) return 0 ;;
+    *) return 1 ;;
+  esac
+}
+
+# Test hook: `DEPLOY_SH_LIB_ONLY=1 source deploy.sh` loads only the helpers above
+# (see test/deploy-helpers.test.sh) — no .env, no required env, no AWS calls.
+# shellcheck disable=SC2317  # the exit is the executed-not-sourced fallback
+if [ "${DEPLOY_SH_LIB_ONLY:-}" = "1" ]; then return 0 2>/dev/null || exit 0; fi
+
 # Load .env
 if [ ! -f .env ]; then
   echo "❌ .env file not found. Run: cp .env.example .env and fill in your values."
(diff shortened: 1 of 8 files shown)


claude-apps-gateway/cdk/scripts/deploy.sh
```diff
@@ -95,9 +95,59 @@ fi
...
+# which is also what preserves TaskSg and any 443 rules you authorized on reused
+# interface endpoints (see NO_ROLLBACK below).
+#
+# The status query itself is a safety control, so it must not fail open: a
+# throttle, expired credential, or network blip would yield an empty status, look
+# like "no stack yet", and run the destructive pass against a live one. Treat ONLY
+# an explicit not-found as absent, and abort on anything else.
+set +e
+STACK_QUERY=$(aws cloudformation describe-stacks --stack-name ClaudeGatewayStack \
+  --query 'Stacks[0].StackStatus' --output text 2>&1)
```
scouturier: If the CLI *succeeds* but emits a stderr warning (e.g. urllib3's `NotOpenSSLWarning` on some macOS installs), the `2>&1` puts it in front of the status, the case falls through to the catch-all, and a `ROLLBACK_COMPLETE` stack that needs pass 1 gets skipped. Fails in the safe direction — but capturing stderr separately would keep the recovery path solid.

claude-apps-gateway/cdk/scripts/deploy.sh
```diff
@@ -95,9 +95,59 @@ fi
...
+fi
+# Running pass 1 is the destructive branch, so IT gets the narrow allowlist: only a
+# stack that is absent or provably holds nothing. Every other status skips to pass 2,
+# which is always the safe default — if the stack is mid-operation or wedged, `cdk
+# deploy` says so itself instead of us enumerating CloudFormation's status list.
+case "$STACK_STATUS" in
+  # No stack, or one holding no resources. ROLLBACK_COMPLETE is a failed FIRST create
+  # (nothing provisioned); CDK deletes and recreates it. REVIEW_IN_PROGRESS is a
+  # change-set-only stack.
+  ""|ROLLBACK_COMPLETE|REVIEW_IN_PROGRESS|DELETE_COMPLETE)
```
scouturier: This decision is pure string→branch logic, which makes it a perfect fit for the repo's stubbed-`aws` shell tests — and CLAUDE.md asks for a case when fixing a deployment trap, and this is the flagship one. Exposing it behind a lib gate like setup.sh's `SETUP_SH_LIB_ONLY` and porting the throwaway matrix from the PR description into `test/deploy-helpers.test.sh` would lock the fix in.

Small aside: `describe-stacks` by *name* never returns `DELETE_COMPLETE` (that needs the stack ID), so that branch is unreachable — harmless, but worth a word so readers don't rely on it.

claude-apps-gateway/cdk/scripts/deploy.sh
```diff
@@ -95,9 +95,59 @@ fi
...
+    echo "Step 1/5: Pass 1 — creating the ECR repository (CDK)..."
+    npx cdk deploy --require-approval never -c imageReady=false "${CDK_CTX[@]}"
+    echo "✅ ECR repository created"
+    ;;
+  *)
+    echo "Step 1/5: stack exists ($STACK_STATUS) — skipping pass 1 (a repo-only deploy would delete the ALB/RDS/service)."
```
scouturier: Dead-end statuses (`*_IN_PROGRESS`, `UPDATE_ROLLBACK_FAILED`, `DELETE_FAILED`, `CREATE_FAILED`) sail through here, run the ~10-minute CodeBuild, then fail at pass 2 anyway. Safe, and I get the "let cdk say so" choice — a cheap bail on `*_IN_PROGRESS|*_FAILED` before the build would just save the wait.

claude-apps-gateway/cdk/scripts/deploy.sh
```diff
@@ -109,36 +159,73 @@ ECR_URI=$(aws cloudformation describe-stacks --stack-name ClaudeGatewayStack \
 # alone, so a name mismatch with the stack can't break the push).
 REPO_NAME="${ECR_URI##*/}"
 echo "   ECR: $ECR_URI (repo: $REPO_NAME)"
+
+# Renaming a deployed gateway in place isn't supported: pass 2 would replace the ECR
+# repo, leaving the service pulling from an empty one. Fail here rather than at a
+# confusing ECS pull error. See cdk/README.md (GATEWAY_NAME).
+if [ "$REPO_NAME" != "$GATEWAY_NAME" ]; then
+  echo "❌ GATEWAY_NAME is '$GATEWAY_NAME' but this stack's repo is '$REPO_NAME'. Set it"
```
scouturier, line 236: Now that pass 1 can be skipped, this branch is reachable with a stack that has no `EcrRepositoryUri` output — `--output text` then yields the literal string `None`, and this message blames a rename that never happened. A quick empty/`None` guard on `ECR_URI` before the comparison would point at the real cause (#288 has the same check — nice convergence point between the two PRs).

claude-apps-gateway/cdk/scripts/deploy.sh
```diff
@@ -109,36 +159,73 @@ ECR_URI=$(aws cloudformation describe-stacks --stack-name ClaudeGatewayStack \
...
 # --- Step 3: Build and push image ---
 echo "Step 3/5: Building and pushing gateway image..."
 
 ACCOUNT_ID=$(aws sts get-caller-identity --query Account --output text)
-BUCKET="claude-gateway-build-${ACCOUNT_ID}"
+# Region-suffixed: CodeBuild requires its S3 source bucket in the PROJECT's region,
+# so the bucket name must vary by region (an account-global name would leave the
+# bucket in the first region and fail a build in any other). See the single-region
+# scope note below.
+BUCKET="claude-gateway-build-${ACCOUNT_ID}-${DEPLOY_REGION}"
```
scouturier, line 250: The rename is right (CodeBuild needs its source bucket in the project's region), but upgraders re-running this strand the old `claude-gateway-build-<acct>` bucket — still holding the stamped config zip — with nothing telling them it's now dead weight. One line in teardown.md would cover it.

claude-apps-gateway/cdk/scripts/deploy.sh
```diff
@@ -109,36 +159,73 @@ ECR_URI=$(aws cloudformation describe-stacks --stack-name ClaudeGatewayStack \
...
+  --assume-role-policy-document '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"Service":"codebuild.amazonaws.com"},"Action":"sts:AssumeRole"}]}' 2>/dev/null || true
 
-  aws iam put-role-policy --role-name claude-gateway-codebuild --policy-name build-perms \
-    --policy-document "{\"Version\":\"2012-10-17\",\"Statement\":[{\"Effect\":\"Allow\",\"Action\":[\"s3:GetObject\",\"s3:ListBucket\"],\"Resource\":[\"arn:aws:s3:::claude-gateway-build-${ACCOUNT_ID}\",\"arn:aws:s3:::claude-gateway-build-${ACCOUNT_ID}/*\"]},{\"Effect\":\"Allow\",\"Action\":[\"ecr:GetAuthorizationToken\"],\"Resource\":\"*\"},{\"Effect\":\"Allow\",\"Action\":[\"ecr:BatchCheckLayerAvailability\",\"ecr:GetDownloadUrlForLayer\",\"ecr:BatchGetImage\",\"ecr:PutImage\",\"ecr:InitiateLayerUpload\",\"ecr:UploadLayerPart\",\"ecr:CompleteLayerUpload\"],\"Resource\":\"arn:aws:ecr:${DEPLOY_REGION}:${ACCOUNT_ID}:repository/${REPO_NAME}\"},{ \"Effect\":\"Allow\",\"Action\":[\"logs:CreateLogGroup\",\"logs:CreateLogStream\",\"logs:PutLogEvents\"],\"Resource\":\"*\"}]}" 2>/dev/null
+aws iam put-role-policy --role-name claude-gateway-codebuild --policy-name build-perms \
```
scouturier, line 289: Now that this runs on *every* deploy, the `2>/dev/null` means a real IAM failure aborts the script with zero output (`set -e`, no `|| true`) — the run just dies after "Step 3/5". `put-role-policy` is quiet on success, so dropping the redirect is enough. (Related: `create-role`'s `|| true` one line up swallows a permissions error, which is exactly what would make this one fail.)

claude-apps-gateway/cdk/scripts/deploy.sh
```diff
@@ -109,36 +159,73 @@ ECR_URI=$(aws cloudformation describe-stacks --stack-name ClaudeGatewayStack \
...
+  --policy-document "{\"Version\":\"2012-10-17\",\"Statement\":[{\"Effect\":\"Allow\",\"Action\":[\"s3:GetObject\",\"s3:ListBucket\"],\"Resource\":[\"arn:aws:s3:::${BUCKET}\",\"arn:aws:s3:::${BUCKET}/*\"]},{\"Effect\":\"Allow\",\"Action\":[\"ecr:GetAuthorizationToken\"],\"Resource\":\"*\"},{\"Effect\":\"Allow\",\"Action\":[\"ecr:BatchCheckLayerAvailability\",\"ecr:GetDownloadUrlForLayer\",\"ecr:BatchGetImage\",\"ecr:PutImage\",\"ecr:InitiateLayerUpload\",\"ecr:UploadLayerPart\",\"ecr:CompleteLayerUpload\"],\"Resource\":\"arn:aws:ecr:${DEPLOY_REGION}:${ACCOUNT_ID}:repository/${REPO_NAME}\"},{ \"Effect\":\"Allow\",\"Action\":[\"logs:CreateLogGroup\",\"logs:CreateLogStream\",\"logs:PutLogEvents\"],\"Resource\":\"*\"}]}" 2>/dev/null
 
-  sleep 10  # Wait for IAM propagation
+# IAM is eventually consistent, so wait AFTER (re)asserting the policy — every run,
+# not only at project creation. A prior fix ran this sleep only when the project
```
scouturier: Tiny: "A prior fix ran this sleep only when…" narrates the script's git history and will read as noise post-merge. The durable constraint — the policy can change on any repo/region change, so wait after every re-assert — is the part worth keeping.

claude-apps-gateway/cdk/scripts/deploy.sh
```diff
@@ -226,8 +313,19 @@ echo ""
...
+# NO_ROLLBACK=1 adds --no-rollback (CloudFormation DisableRollback). Needed when
+# reusing a VPC whose interface endpoints don't yet allow 443 from the task SG: the
+# service can't stabilize, and the default rollback deletes the just-created TaskSg
+# — the very SG you have to authorize on those endpoints. Retaining it makes the
+# documented deploy → authorize → redeploy sequence possible. Off by default: the
+# failure leaves the stack in UPDATE_FAILED, which you then have to either retry or
+# abandon by hand (`npx cdk rollback` / `aws cloudformation rollback-stack` —
+# continue-update-rollback does NOT apply, it only takes UPDATE_ROLLBACK_FAILED).
+CDK_DEPLOY_FLAGS=(--require-approval never)
+[ "${NO_ROLLBACK:-0}" = "1" ] && CDK_DEPLOY_FLAGS+=(--no-rollback)
```
scouturier, line 418: Scope call for the maintainers rather than a defect: this automates a one-time recovery whose by-hand command is already documented in the section it serves, so the repo's fix-vs-feature default would trim it to the doc note — though I see the value (hand-running pass 2 means reconstructing all the `-c` context from `.env` mid-recovery). If it stays, one caveat worth adding: CDK refuses `--no-rollback` when an update requires resource replacement, so a stray exported `NO_ROLLBACK=1` on a later update fails confusingly.

claude-apps-gateway/docs/deployment.md
```diff
@@ -225,6 +238,49 @@ To roll a new image later: push a new tag, then re-run pass 2 with
...
+`-c vpcId=vpc-...` for CDK — typically to keep an existing Client VPN association
+intact. Three things to know.
+
+**CIDR.** `setup.sh` creates subnets at fixed CIDRs (`10.20.0.0/24`, `10.20.1.0/24`,
+`10.20.10.0/24`, `10.20.11.0/24`), so the VPC must be `10.20.0.0/16` or a superset
+with those four ranges free; anything else fails at `create-subnet`. The CDK track
+imports the VPC's existing subnets instead, so any CIDR works.
+
+**Endpoint security groups.** If the VPC already has the Bedrock / Secrets Manager /
+ECR / CloudWatch **interface** endpoints, set `CREATE_VPC_ENDPOINTS=false` (or
```
scouturier: Small track mix-up: `CREATE_VPC_ENDPOINTS` only exists on the CDK track — setup.sh has no such flag (it adopts pre-existing endpoints automatically, describe-before-create). One clause would save a Track-A reader from setting a no-op env var.

claude-apps-gateway/docs/deployment.md
```diff
@@ -225,6 +238,49 @@ To roll a new image later: push a new tag, then re-run pass 2 with
...
+
+**Ordering, which differs by track.** `setup.sh` creates its task SG (`$P-task-sg`)
+before tasks start: authorize it, then re-run. CDK's `TaskSg` exists only in pass 2,
+so it can't be preauthorized — and a pass 2 that can't stabilize normally rolls back
+and *deletes* the SG you were about to authorize. Deploy that pass with rollback
+disabled so it survives:
+
+```bash
+npx cdk deploy --no-rollback ...   # by hand
+NO_ROLLBACK=1 ./scripts/deploy.sh  # or via deploy.sh
```
scouturier, line 274: This (rightly) recommends deploy.sh — but the NOTE ~30 lines up still says deploy.sh "does not pass CDK context, so this guide — not that script — is the supported route", which is no longer true either (deploy.sh builds the full `CDK_CTX` from `.env`). Worth reconciling the NOTE while you're in the file.

claude-apps-gateway/docs/teardown.md
```diff
@@ -120,17 +120,53 @@ aws ecr delete-repository --repository-name "$P" --force --region "$AWS_REGION"
...
+
 ```bash
-VPC_ID=$(aws ec2 describe-vpcs --filters "Name=tag:Project,Values=$P" \
-  --query 'Vpcs[0].VpcId' --output text --region "$AWS_REGION")
+# Record reuse intent BEFORE the lookup fills VPC_ID in: if you set VPC_ID yourself,
+# the VPC is yours to keep, whatever tags it carries. (A pre-existing VPC can legally
+# be tagged Project=claude-gateway — e.g. you built it for this project and put the
+# Client VPN in it — so the tag alone does NOT prove setup.sh created it.)
+VPC_REUSED=0
+[ -n "${VPC_ID:-}" ] && VPC_REUSED=1
```
scouturier: This block is stateful across re-pastes: the lookup below fills `VPC_ID` into the shell, so re-running it (which the SG-straggler note anticipates) flips `VPC_REUSED=1` on a setup.sh-*created* VPC and silently skips `delete-vpc` with the "supplied for reuse" message. A separate sentinel, or `unset VPC_ID` at the end of the block, fixes it.

claude-apps-gateway/docs/teardown.md
```diff
@@ -139,36 +175,78 @@ if [ -n "$NAT" ] && [ "$NAT" != "None" ]; then
...
+aws ec2 describe-nat-gateways --filter "Name=vpc-id,Values=$VPC_ID" --output table --region "$AWS_REGION" \
+  --query "NatGateways[].[NatGatewayId,State,Tags[?Key=='Project']|[0].Value]"
+```
+
+A `Project=$P` tag is a hint, not proof: `setup.sh` adopts a pre-existing subnet,
+IGW, NAT, or SG that matches what it wants rather than creating one, and tags only
+resources it creates — so anything it adopted still carries your tags. Delete what
+you know you added; leave the rest. The four subnets `setup.sh` creates are
+`$P-public-a/b` and `$P-private-a/b`; its SGs are `$P-alb-sg`, `$P-task-sg`, and
+`$P-rds-sg`.
```
scouturier: setup.sh also creates `$P-vpce-sg` (`ensure_sg` at setup.sh:458) — worth adding here so the hand-teardown list is complete.

scouturier (commented):
Ran a deep pass on this — shell semantics (on a real macOS bash 3.2.57), the CloudFormation/CDK/CodeBuild claims, docs vs code, and the repo's sample-scope lens. The core fix is exactly right: fail-closed status query, allowlist on the destructive branch, and it protects what the gateway ops doc makes the *routine* operation (binary patch cadence, config changes, and secret rotation all re-run this script) — so the skip-pass-1 branch is the hot path, which is why a few of the inline notes below weigh error-reporting more than usual. Both shell suites pass on the branch (19 + 9).

Inline comments are anchored where each item lives. Two things had no line to sit on:

- **PR body drift** — the section-5 word-count figures don't match any commit on main (the four table cells actually grew), and "setup-helpers: 14 passed" is 19 on this branch. Worth syncing before squash-merge since the body becomes the commit message.
- The one blocking-by-convention ask (a committed test for the status gate) is inline at the case statement.

Everything else verified out — the teardown tag-scoping in particular is nicely done: every filter matches a tag setup.sh actually stamps.

bluedoors:
Thanks — that was a genuinely useful pass. Pushed as 2b928b6. Eleven of the twelve inline items are in; one I'd like to push back on, and one is a scope call for you.

### Push-back: the cheap bail on dead-end statuses

I implemented this first, then went and read the pinned CDK (2.1129, from `cdk/node_modules/aws-cdk`) to write the comment — and the premise doesn't hold. Three of the four statuses you listed are recoverable by `cdk deploy`, not dead ends:

| Status | What `cdk deploy` does | Where |
|---|---|---|
| `*_IN_PROGRESS` | **waits** for the operation to settle, then deploys | `stabilizeStack`, `index.js:214459` |
| `ROLLBACK_COMPLETE`, `ROLLBACK_FAILED` | **deletes and recreates** the stack | `isCreationFailure`, `index.js:212446` + `227709` |
| `CREATE_FAILED`, `UPDATE_FAILED`, `UPDATE_ROLLBACK_FAILED` | fail-paused → **asks** to roll back, then retries | `isRollbackable` → `failpaused-need-rollback-first`, `index.js:228093` |

So my bail refused deploys CDK completes on its own — on `ROLLBACK_FAILED` and on every `*_IN_PROGRESS`, turning a case that heals itself into a hard failure. I've reverted it; `needs_pass_one` is behaviourally identical to what you reviewed. The original comment ("if the stack is mid-operation or wedged, `cdk deploy` says so itself instead of us enumerating CloudFormation's status list") turns out to have been the right call, so I've kept that design and written the evidence into the comment at the gate instead.

**Your instinct was pointing at something real, though**, just not where either of us thought. On a fail-paused stack CDK asks for confirmation — and `--require-approval never` does **not** waive it (that flag only covers security-group/IAM approval; only `--force` does). So an interactive run stops for a y/n and a non-TTY run dies with `TtyNotAttached` (`cli-io-host.js:632`) — in both cases after the ~10-minute build, which is exactly the wait you were trying to save. A correct bail would need status × `NO_ROLLBACK` × predicting whether the change set contains a replacement, which is more machinery than this example wants, so it's documented at the gate rather than automated. Happy to add it if you'd rather have the check.

### Scope call for you: `NO_ROLLBACK`

Left in, with your caveat added — and it belongs in the script rather than only the docs, because the trap is a *leftover* export: CDK refuses an update requiring resource replacement while rollback is disabled (`cdk-toolkit.js:1501`; also called out in `--rollback`'s own help text), so a stale `NO_ROLLBACK=1` fails an unrelated later change for a reason that has nothing to do with it. Both the comment and `docs/deployment.md` now say to set it per run. Still your call whether the flag earns its place at all.

### The rest

- **Test for the gate** — `test/deploy-helpers.test.sh`, behind a `DEPLOY_SH_LIB_ONLY` gate mirroring `SETUP_SH_LIB_ONLY`. 31 assertions covering the allowlist, the fail-closed query, and the stderr case below. Also noted in the comment that the `DELETE_COMPLETE` arm is unreachable via describe-by-name, as you said.
- **stderr in the status** — fixed by capturing stdout and stderr separately, and it's asserted: a `NotOpenSSLWarning` alongside a successful query now leaves `ROLLBACK_COMPLETE` recoverable. Worth noting this compounded with the `EcrRepositoryUri` item below — the contaminated status skipped pass 1, and Step 2 then reported it as a rename.
- **`EcrRepositoryUri` → `None`** — guarded. Slight refinement on the mechanism: `--output text` prints `None` when `Outputs` is absent (null), but an *unmatched filter* gives an empty string, so the guard checks both. The message says explicitly that it is not a `GATEWAY_NAME` problem.
- **`put-role-policy 2>/dev/null`** — dropped. Took your related aside too: `create-role`'s `|| true` is narrowed to `EntityAlreadyExists`, so the permissions error that would cause the `put-role-policy` failure now surfaces with the API's own message.
- **Stranded build bucket** — `docs/teardown.md` gained a CodeBuild-artifacts section (role, project, region-suffixed bucket, and the unsuffixed one an upgrade leaves behind). It also holds the staged `claude` binary, so it's ~100 MB of dead weight, not just the config.
- **`CREATE_VPC_ENDPOINTS` track mix-up** — fixed, with the reason: `setup.sh` describes before creating (`setup.sh:474`), so it adopts existing endpoints and needs no flag.
- **The stale `NOTE`** — reconciled, and there were three more copies of the same claim in `cdk/README.md`. The real remaining deltas are the unverified staged binary and the `:latest` tag.
- **Teardown re-paste flipping `VPC_REUSED`** — the flag is now decided once, on the first paste, so a re-run keeps the answer instead of re-deriving it from the now-populated `VPC_ID`. Chose that over `unset VPC_ID` so it doesn't clobber a value you exported.
- **`$P-vpce-sg`** — added, with the note that it exists even when the endpoints themselves were adopted, so it may be unused.
- **Git-history narration in the sleep comment** — gone, constraint kept.
- **PR body drift** — you were right, and by more than the figures suggested: measured against `main` the four cells *grew* (40→54, 21→40, 28→46, 27→47). The "111 → 54" numbers were against a discarded draft of this branch. The body now says that plainly, and the test counts are corrected (19 setup-helpers, 9 stamp-config, 31 deploy-helpers, 17 CDK) — including dropping the "no new test files" line, which is no longer true.

While in the files I also fixed two stale anchors: `#regions` in the `DEPLOY_REGION` row (should be `#regions--data-residency`, as its sibling rows use) and `cdk/README.md`'s binary-staging link. All three shell suites also pass under macOS's system `/bin/bash` 3.2.57, matching your setup.

scouturier:
Re-reviewed 2b928b6. All eleven addressed items are in the code, the four suites pass here under bash 3.2.57 (31 / 19 / 9 / 17), and every new anchor resolves. I also checked the CDK behaviour claims in the gate comment against the CLI the lockfile installs (2.1133.0): the `*_IN_PROGRESS` wait, the `ROLLBACK_COMPLETE` / `ROLLBACK_FAILED` delete-and-recreate, and the fail-paused prompt that `--require-approval never` does not waive all hold. Reverting the bail was the right call.

Three edits before merge, all in the new revision:

- **`CLAUDE.md:99`** says the suite covers "early bail on a wedged stack". There is no such test; section 3 of `deploy-helpers.test.sh` asserts the opposite.
- **`deploy.sh:184` and `deploy-helpers.test.sh:106`** cite "pinned CDK (2.1129)". The lockfile pins 2.1133.0, and the number will drift again on the next bump. Point at the source instead ("the CDK CLI pinned in `cdk/package-lock.json`") and, if you want the dated observation, add a line to `docs/upstream-watch.md` where the sample already keeps its "Validated on" notes.
- **`docs/deployment.md:277-290`**: the retry after authorizing 443 needs `NO_ROLLBACK=1` again. The stack is `UPDATE_FAILED`, which CDK treats as fail-paused, so a plain `./scripts/deploy.sh` hits the roll-back-first prompt (or `TtyNotAttached` non-interactively) after the image build, and answering "y" rolls back and removes the `TaskSg` you just authorized. One clause: retry with `NO_ROLLBACK=1` too; plain runs once the stack is back to `UPDATE_COMPLETE`.

Two optional nits: `deploy.sh:196`, `--yes` waives that prompt as well as `--force`. `deployment.md:288`, it is the CDK CLI that intercepts a replacement under `--no-rollback` and prompts to re-enable rollback, not CloudFormation refusing.

Scope call: keep `NO_ROLLBACK`. Two lines, off by default, and the by-hand alternative means rebuilding eight `-c` values from `.env` mid-recovery.

Thanks for the thorough turnaround on this one.

claude-apps-gateway/cdk/scripts/deploy.sh
```diff
@@ -226,8 +400,24 @@ echo ""
...
+# failure leaves the stack in UPDATE_FAILED, which you then have to either retry or
+# abandon by hand (`npx cdk rollback` / `aws cloudformation rollback-stack` —
+# continue-update-rollback does NOT apply, it only takes UPDATE_ROLLBACK_FAILED).
+#
+# Set it per-run (`NO_ROLLBACK=1 ./scripts/deploy.sh`), not exported in your shell or
+# .env: CloudFormation REFUSES an update that requires replacing a resource while
+# rollback is disabled, so a leftover NO_ROLLBACK=1 makes an unrelated later change
+# fail for a reason that has nothing to do with the change.
+CDK_DEPLOY_FLAGS=(--require-approval never)
+[ "${NO_ROLLBACK:-0}" = "1" ] && CDK_DEPLOY_FLAGS+=(--no-rollback)
```
scouturier, line 418: This is the only new capability in the PR, and config already reaches the same end state. The endpoint SG rule does not have to reference `TaskSg`: authorizing 443 from the **VPC CIDR** needs nothing to exist first, lets the initial pass 2 stabilize, and can be tightened to `TaskSg` afterwards. Dropping the flag also drops the `UPDATE_FAILED` recovery path and the "never export it" footgun that need documenting in two files.

claude-apps-gateway/docs/deployment.md
```diff
@@ -219,11 +232,65 @@ To roll a new image later: push a new tag, then re-run pass 2 with
...
+ECR / CloudWatch **interface** endpoints, tell the CDK track not to recreate them with
+`CREATE_VPC_ENDPOINTS=false` (`deploy.sh`) or `-c createVpcEndpoints=false` (by hand)
+— AWS allows one private-DNS endpoint per service per VPC. `setup.sh` has no such
+flag and needs none: it describes before creating, so it adopts endpoints already in
+the VPC. Neither track then touches those endpoints' security groups, so **you** must
+allow 443 from the gateway task SG on each, or tasks time out fetching secrets,
+images, and logs. S3 is a *gateway* endpoint: no SG, no 443 — it just needs an
+association with the tasks' private route table.
+
+**Ordering, which differs by track.** `setup.sh` creates its task SG (`$P-task-sg`)
```
scouturier, line 266: Same point as the `deploy.sh` comment: the deploy → authorize → redeploy sequence works without disabling rollback if the endpoint SGs allow 443 from the VPC CIDR for the first pass 2, then get tightened to `TaskSg`. That replaces this subsection, the code path, and the `UPDATE_FAILED` paragraph below with about three lines of prose.

claude-apps-gateway/docs/deployment.md
```diff
@@ -219,11 +232,65 @@ To roll a new image later: push a new tag, then re-run pass 2 with
...
+with those four ranges free; anything else fails at `create-subnet`. The CDK track
+imports the VPC's existing subnets instead, so any CIDR works.
+
+**Endpoint security groups.** If the VPC already has the Bedrock / Secrets Manager /
+ECR / CloudWatch **interface** endpoints, tell the CDK track not to recreate them with
+`CREATE_VPC_ENDPOINTS=false` (`deploy.sh`) or `-c createVpcEndpoints=false` (by hand)
+— AWS allows one private-DNS endpoint per service per VPC. `setup.sh` has no such
+flag and needs none: it describes before creating, so it adopts endpoints already in
+the VPC. Neither track then touches those endpoints' security groups, so **you** must
+allow 443 from the gateway task SG on each, or tasks time out fetching secrets,
```
scouturier, line 262: This added sentence attributes the timeout to the endpoint SGs alone, and the new listing block below doesn't ask the reader to check private DNS on the endpoints being reused. Same symptom, different cause, so someone following this section audits SGs and finds nothing wrong. One clause here plus `describe-vpc-endpoints --query 'VpcEndpoints[].[ServiceName,PrivateDnsEnabled,SubnetIds]'` in the listing block would make the new section complete.

claude-apps-gateway/cdk/scripts/deploy.sh
```diff
@@ -95,50 +150,169 @@ fi
...
+STACK_STATUS="$(classify_stack_query "$STACK_RC" "$STACK_OUT" "$STACK_ERR")" || {
+  echo "❌ Could not determine the ClaudeGatewayStack status, so deploy.sh cannot"
+  echo "   tell a first deploy from an existing one — refusing to continue, because"
+  echo "   guessing wrong runs a repo-only deploy that would delete the RDS instance"
+  echo "   and the rest of the running stack. Fix the AWS call and re-run:"
+  echo "   $STACK_ERR"
+  exit 1
+}
+
+# Only the pass-1 decision is made here. Every other status is CDK's to report — this
```
scouturier, line 182: 123 of the 207 lines this PR adds to `deploy.sh` are comments or blanks, taking the file 285 → 475 lines. The rationale is worth keeping, but this 24-line account of CDK 2.1129's status handling reads as documentation, and the script's job in this repo is to be readable as a worked example. Suggest moving it next to the pass-1 `[!CAUTION]` in `deployment.md` and leaving a one-line pointer here, the same split the PR already uses for the reuse guidance. `teardown.md:126-146` and `:148-157` have the same shape.

scouturier (commented):
Reviewed at the PR head. Verified rather than assumed: all three shell suites pass on `/bin/bash` 3.2.57 (31 / 19 / 9), `bash -n` clean on both scripts, the `[ ... ] && ARR+=(...)` line does not abort under `set -e` (checked empirically), every new anchor resolves, and each teardown tag filter matches something `setup.sh` actually stamps in its create branch only. `main:deploy.sh:99` does run pass 1 with no status guard, so the data-loss bug is real. Description matches the diff, including the "31 passed" and "three more copies" claims.

Fix-vs-feature triage: the pass-1 gate, endpoint tagging, tag-scoped teardown, region-suffixed bucket, `update-project` source assertion, IAM re-assert and `GATEWAY_NAME` refusal are all fixes and I would land them as they are. Three comments inline: one new capability that config already reaches, one gap in the new reuse guidance, and commentary volume.

**Target end-state:** land the fixes as they are; drop `NO_ROLLBACK` in favour of a CIDR-then-tighten note; add private DNS to the reuse checklist; move the CloudFormation-status rationale into `deployment.md`.

scouturier:
Thanks for the rebase and the FAULT/TIMED_OUT fix. I re-ran the suites at `7384fc7` (31 / 19 / 9 on bash 3.2.57) and did a full pass over the scripts. No new issues. Two things remain, and then I'll approve and merge:

1. **Drop `NO_ROLLBACK`** (`deploy.sh:405-422`, `docs/deployment.md:266-290`, plus the references at `deploy.sh:161`, `:197-198` and `deploy-helpers.test.sh:88`). This reverses my 09-03 call, sorry for the back-and-forth. Allowing 443 from the VPC CIDR on the reused endpoints' SGs before the first deploy lets pass 2 stabilize with rollback on, so neither the flag nor its recovery path is needed. Suggested replacement for the Ordering subsection:
   > **Ordering.** The task SG doesn't exist until the stack creates it (CDK's `TaskSg` only appears in pass 2), and a pass 2 that can't reach the endpoints rolls back and deletes it. So allow 443 from the **VPC CIDR** on the interface endpoints' SGs before the first deploy. Once the stack is up you can tighten each rule to the task SG (`TaskSg`, or `$P-task-sg` for `setup.sh`) and drop the CIDR rule.
2. **`CLAUDE.md:139`**: remove "early bail on a wedged stack". The suite asserts the opposite.

I'll tidy the squash commit message at merge.

bluedoors:
All three are in at `1cd365f` — `NO_ROLLBACK` dropped for the VPC-CIDR-then-tighten note, `CLAUDE.md:139` corrected, PR body resynced — and I moved the CDK status commentary out of `deploy.sh` into `docs/deployment.md` while I was there (475 → 457 lines). Suites still 31 / 19 / 9, and `npm test` is 18, not 17: two of its assertions had been failing locally against stale gitignored `tsc` output shadowing the `.ts`, nothing in the branch.

scouturier (approved):
Verified at `1cd365f`: `NO_ROLLBACK` gone, `CLAUDE.md` corrected, body in sync with the diff. Shell suites 31 / 19 / 9 on bash 3.2.57, build check green. Thanks for seeing this one through.

PR #269, opened 2026-07-27
feat: add claude-code-on-lambda-microvm — private Claude Code and VS Code dev environments on AWS Lambda MicroVMs

roshansthomas:
## Summary

Adds **`claude-code-on-lambda-microvm`**: private, per-developer Claude Code and VS Code development environments on AWS Lambda MicroVMs. One CDK stack (`ClaudeMicrovmStack`) gives each developer a disposable Linux ARM64 workspace where Claude Code, VS Code Server, terminals, and language tooling all execute inside the MicroVM — the laptop is only the user-interface endpoint, and inference stays on a private Amazon Bedrock Runtime endpoint by default.

## Why this is a strong addition to the repo

**It answers the question enterprises are actually asking.** Security teams increasingly will not allow an autonomous coding agent to run on corporate laptops with broad filesystem and credential access. This sample shows the alternative: Claude Code runs in an isolated, disposable, service-managed MicroVM with scoped VPC egress, temporary execution-role credentials, and encrypted workspace checkpoints — while the developer keeps a native local experience (their own terminal, or VS Code Desktop via outbound Microsoft Remote Tunnels).

**It is the first Lambda MicroVM sample in the repo** and complements the existing catalog without overlapping it: `claude-code-on-agentcore` covers headless/autonomous execution, `claude-apps-gateway` covers governed inference for local clients; this covers **interactive remote development** — and it composes with both (the gateway is a retained legacy inference mode, and AgentCore Gateway is the optional governed MCP tool path).

**Defense-in-depth is designed in, not bolted on:**
- Private REST API (execute-api VPC endpoint) with Cognito-authorizer (portal) or SigV4 (CLI) auth; disjoint owner namespaces prevent cross-channel session takeover
- No SSH daemon, no public IP, no bastion, no inbound ports — VS Code connects through outbound-only Microsoft dev tunnels
- Bedrock inference over a VPC endpoint using MicroVM execution-role credentials; no long-lived secrets in the image
- Root-owned, machine-managed Claude provider configuration the developer cannot override
- KMS-encrypted, versioned S3 workspace checkpoints; tunnel credentials and device tokens are deliberately ephemeral
- cdk-nag (AwsSolutions) checks wired into the build

**It handles the hard operational realities of the MicroVM platform** rather than demoing around them: an eight-hour invocation limit (a per-minute reconciler checkpoint-terminates early and restores `/workspace` into a fresh MicroVM), idle suspend/resume policies, orphan sweeping, and atomic owner/workspace claims that prevent duplicate sessions.

**No local Docker anywhere** — the MicroVM image is built service-side from a plain source ZIP, and deploy/provision/client/cleanup scripts run on macOS, Linux, and stock Windows.

## What is included

- `infra/` — one CDK stack: private VPC, private REST API, control Lambda, DynamoDB session state, KMS-encrypted S3 checkpoints, optional Cognito portal, optional sample Client VPN
- `control-plane/` — lifecycle service (create/suspend/resume/terminate, ownership, tunnel auth)
- `microvm/` — the in-VM Python agent (provider config, checkpoints, VS Code tunnel supervision) and the AgentCore MCP bridge
- `client/` — laptop CLI for terminal attach and VS Code session bootstrap
- `portal/` — optional Cognito-backed browser portal for environment lifecycle
- `docs/deployment-guide.md` — full architecture and operations reference with an explicit default-mode deployment path; all legacy-gateway content clearly marked and skippable
- Architecture diagram (Draw.io source + PNG)

## Validation

- 178 tests pass: 126 vitest (client, control plane, portal, infra, scripts) + 52 Python unittest (MicroVM agent)
- `tsc` clean for the root project and the `microvm/` package; `cdk synth` passes with cdk-nag checks
- Deployed and validated end to end in Bedrock mode: Cognito portal sign-in, environment creation, and a live terminal Claude Code session on Bedrock. VS Code mode has completed Microsoft Remote Tunnels registration from a live MicroVM; editor acceptance on macOS/Windows is documented as not yet exercised
- No account identifiers or engagement-specific values in the tree; `deployment.example.json` ships with safe defaults (`inferenceMode: bedrock`, `createClientVpn: false`, `enablePortal: false`)

## Limitations (documented in the README)

Single-Region sample; NAT egress is any-IPv4-HTTPS (production allowlisting via Network Firewall is called out); Microsoft dev-tunnels relay requires enterprise approval; the shell WebSocket framing is validated against the live service since AWS does not yet document the frame details.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


claude-code-on-lambda-microvm/infra/lib/nag-acks.ts
```diff
@@ -0,0 +1,301 @@
...
+    id: 'AwsSolutions-IAM4[Policy::arn:<AWS::Partition>:iam::aws:policy/service-role/AmazonAPIGatewayPushToCloudWatchLogs]',
+    reason:
+      'AWS-managed policy required by API Gateway account settings to ' +
+      'deliver execution and access logs to CloudWatch.',
+  });
+
+  // --- MicroVM execution role ---------------------------------------------
+
+  ack('MicrovmExecutionRole/DefaultPolicy/Resource', {
+    id: 'AwsSolutions-IAM5[Resource::arn:<AWS::Partition>:logs:us-east-1:<AWS::AccountId>:log-group:/<ProjectName>/microvms:*]',
```
scouturier: **Blocking: this hardcoded `us-east-1` makes `cdk synth` fail in every other region.**

This ID (and the ones at lines 163, 197, 202, 270) embeds a literal region, but the region is configurable — `infra/bin/app.ts:12` reads it from context and `scripts/deploy.ts:68` forwards it from `deployment.json`. `docs/deployment-guide.md:964` actively tells readers to set `"region": "<region>"`.

Confirmed locally: `npx cdk synth -c region=eu-west-1` exits 1 with 10 unacknowledged `AwsSolutions-IAM5` findings, cdk-nag asking for `logs:eu-west-1:…`, `execute-api:eu-west-1:…`, `ec2:eu-west-1:…`.

Building these from `Stack.of(this).region` would fix it — the same way `<AWS::AccountId>` is already tokenized here.

claude-code-on-lambda-microvm/infra/lib/platform-stack.ts
```diff
@@ -0,0 +1,1208 @@
...
+      inferenceMode !== 'claude-gateway' &&
+      inferenceMode !== 'bedrock'
+    ) {
+      throw new Error(
+        'inferenceMode context must be claude-gateway or bedrock',
+      );
+    }
+    const bedrockModelId =
+      this.node.tryGetContext('bedrockModelId') ??
+      'us.anthropic.claude-sonnet-4-6';
```
scouturier: **Blocking: default model is a previous generation, and the check below rejects the current ones.**

`us.anthropic.claude-sonnet-4-6` is now listed under Legacy in the models overview; current is Sonnet 5 / Opus 5 / Fable 5. Same value is duplicated in `deployment.example.json:9` and `microvm/agent.py:75-77`.

Upgrading isn't a one-liner, which is why I'm flagging it rather than suggesting a swap: current Bedrock IDs use the `anthropic.` prefix and are served via the Messages-API endpoint (`CLAUDE_CODE_USE_MANTLE=1`), so the regex on line 57 rejects them outright and `claude_provider_environment` (`agent.py:1140-1151`) has to change with it.

Pinning a version is correct and doc-endorsed — it's the generation that's stale.

claude-code-on-lambda-microvm/infra/lib/platform-stack.ts
```diff
@@ -0,0 +1,1208 @@
...
+    if (
+      inferenceMode === 'bedrock' &&
+      !/^(?:us|global)\.anthropic\.claude-[A-Za-z0-9._:-]{1,180}$/.test(
```
scouturier: This pattern only admits `us.`/`global.` inference-profile IDs, so any current-generation model ID (`anthropic.claude-sonnet-5`) throws `bedrockModelId is not an approved Claude model ID`.

That constraint isn't documented anywhere: the deployment guide only ever says "an approved Bedrock model or inference profile" (lines 1037, 1165, 1185; the worksheet at 300 just says "Approved values"), and this error isn't in the troubleshooting table. A reader who sensibly tries a newer model gets a hard failure with nothing to consult.

A short "choosing a model ID" section covering the two endpoint families would help.

claude-code-on-lambda-microvm/control-plane/src/service.ts
```diff
@@ -0,0 +1,905 @@
...
+      }
+      recordStored = true;
+
+      const checkpoint = await this.options.checkpoints.createAccess(
+        ownerHash,
+        workspaceId,
+      );
+      // Keep existing image-19 terminal sessions deployable while the feature
+      // image is building. VS Code and Claude.ai require the v3 contract.
+      const payloadVersion =
```
scouturier: **Blocking (quick fix): leftover from the deployment this was built for.**

The comment above — "Keep existing image-19 terminal sessions deployable while the feature image is building" — describes your private environment's rollout state. A brand-new sample has no previously deployed sessions to stay compatible with, so there's nothing for v2 to serve.

Collapsing to v3 unconditionally lets you delete this branch, the `payloadVersion === 3` conditional below, and the matching gate at `microvm/agent.py:897`.

claude-code-on-lambda-microvm/microvm/agent.py
```diff
@@ -0,0 +1,1552 @@
...
+            raise ValueError("Invalid tunnelName")
+    elif tunnel_name is not None:
+        raise ValueError("tunnelName requires vscode accessMode")
+    if payload_version == 2 and (
+        access_mode_value != "terminal"
+        or inference_mode == "claude-ai"
+        or tunnel_name is not None
+    ):
+        raise ValueError(
+            "Run hook payload version 3 is required for this session"
```
scouturier: Other half of the v2/v3 split from `control-plane/src/service.ts:194`. Once the control plane always sends v3, this check and the version branching around it can go.

claude-code-on-lambda-microvm/control-plane/src/tunnel-auth-worker.ts
```diff
@@ -0,0 +1,651 @@
...
+  device?: {
+    verificationUri: string;
+    userCode: string;
+  };
+  starting: boolean;
+  ready: boolean;
+  exitCode?: number;
+}
+
+export class TunnelOutputParser {
```
scouturier, line 60: **Session startup depends on matching text the VS Code CLI prints.**

This pattern matches a string emitted by `microvm/vscode_tunnel.py:631-637`, and the same regex is duplicated in `client/src/cli.ts:264-276`. If Microsoft rewords that line, tunnel startup breaks silently — and it has to be fixed in two places that can drift apart.

Worth hoisting into one shared exported constant, with a comment marking it as a workaround for the current CLI output (`README.md:276-279` already makes this admission for the WebSocket framing — same treatment fits here).

claude-code-on-lambda-microvm/infra/lib/nag.ts
```diff
@@ -0,0 +1,26 @@
...
+ * ACKNOWLEDGED_RULES_METADATA_KEY (walking up the construct tree), so writing
+ * the same metadata directly is equivalent — same audit trail in the
+ * synthesized template, no ID-format restriction.
+ */
+export function ackNag(
+  scope: IConstruct,
+  ...rules: Array<{ id: string; reason: string }>
+): void {
+  for (const rule of rules) {
+    scope.node.addMetadata(
```
scouturier: **This reaches past a public CDK API, and a clean `cdk synth` depends on it.**

The docstring is honest about why: `Validations.of(x).acknowledge()` rejects granular IDs like `AwsSolutions-IAM5[Resource::*]` because of the `::`. Writing the metadata by hand does work today, and the reasoning is sound.

The risk is that `ACKNOWLEDGED_RULES_METADATA_KEY` isn't a supported contract, so a CDK upgrade could break the build with no warning — and because every suppression in `nag-acks.ts` routes through this helper, it'd break all of them at once.

cdk-nag's `NagSuppressions.addResourceSuppressions` with `appliesTo` handles granular findings through a supported path. Worth filing the `acknowledge()` limitation upstream as well.

claude-code-on-lambda-microvm/client/src/vscode.ts
```diff
@@ -0,0 +1,384 @@
...
+import { posix, win32 } from 'node:path';
+import process from 'node:process';
+
+const TUNNEL_NAME_PATTERN = /^[A-Za-z0-9-]{1,20}$/;
+export const REMOTE_TUNNELS_EXTENSION =
+  'ms-vscode.remote-server';
+export const REMOTE_TUNNELS_EXTENSION_VERSION =
+  '1.6.2026061009';
+export const REMOTE_TUNNELS_EXTENSION_SPEC =
+  `${REMOTE_TUNNELS_EXTENSION}@${REMOTE_TUNNELS_EXTENSION_VERSION}`;
```
scouturier, line 24: **Pre-release pin and an auth-broker override, presented as permanent config.**

This pins a pre-release build of `ms-vscode.remote-server` (installed with `--pre-release --force` in `remoteTunnelsInstallArguments`), and `VSCODE_COMPATIBILITY_SETTINGS` below disables the Microsoft auth broker via `microsoft-authentication.implementation: msal-no-broker`.

Both are workarounds for how those components behave right now. The problem is that `docs/deployment-guide.md`'s troubleshooting table lists the broker setting as ordinary configuration, so a reader has no way to know it's temporary or when to stop applying it.

Worth labeling both as current-behavior workarounds and naming what would supersede them. Also note the PR body says macOS/Windows editor acceptance isn't exercised yet — that's the path these two settings most affect.

scouturier (commented):
Solid, security-conscious build. I verified the PR body's claims at `7ca3501`: `tsc` clean, 178 tests pass (126 vitest + 52 Python), `cdk synth` passes cdk-nag in `us-east-1`.

I've left specific notes inline. Two of them block merge in my view — the region-locked cdk-nag IDs (synth fails outside `us-east-1`, which the docs actively tell readers to try) and the legacy default model whose validation rule rejects current model IDs. The `image-19` leftovers are a quick delete.

Three things don't belong on any single line, so they're here:

**~390 lines re-document the existing `claude-apps-gateway` sample.** 26 of the 60 headings in `docs/deployment-guide.md` are `(legacy gateway mode)` — Entra setup, the two-pass CodeBuild flow, even the same `ClaudeGatewayStack -c imageReady=true` command that sample's own teardown doc uses. Two copies will drift apart, and it buries the primary Bedrock path readers actually want. I'd link out instead. This is the fourth thing I'd want resolved before merge, since it only gets harder to unwind later.

**The VPN and portal stretch the sample past what it's teaching.** The optional Client VPN (`platform-stack.ts:299-405`) is ordinary corporate networking, shipping a CA valid for ten years with no revocation path — fine for a demo, but a lot of surface area in a sample about MicroVMs, and it reads better as an optional doc than as deployable code. The Cognito portal (`platform-stack.ts:838-1074` plus a 1099-line `portal/site.ts`) is close to a finished product UI, and it can't be used alongside the CLI at all: the portal identifies owners as `oidc:<sub>` while the CLI uses a hash of the IAM ARN, so as `README.md:302-306` states outright, the CLI cannot attach to a portal-created environment. A reader who follows the portal path and then wants a terminal has to start over.

**Editor traffic depends on a third-party relay.** All source, terminal, and editor-protocol traffic is relayed through Microsoft's dev-tunnels service (`README.md:284-286` is upfront that this needs data-residency and TLS-inspection sign-off). Worth stating in the deployment guide's prerequisites too, not only in limitations — it's an approval a reader needs *before* they start.

**Smaller things:** version numbers are duplicated between `microvm/Dockerfile:3-7` and `scripts/provision-microvm.ts:53-58`; the default tunnel provider disagrees between `control-plane/src/service.ts:89` (`github`) and `client/src/cli.ts:313` (`microsoft`); and `listForOwner` scans the whole table then filters in memory (`service.ts:126-132`) where a GSI query would do.

Happy to re-review once the blocking items are in.

roshansthomas:
Thanks for the detailed review. Addressed in `497fcb1`:

- Made cdk-nag suppressions region-aware and moved them to the public `NagSuppressions` API. Default and `eu-west-1` synths now pass.
- Updated the default to Sonnet 5, added Mantle Messages API support, and documented model ID versus inference profile endpoint families.
- Removed the image-19 payload compatibility path. The control plane now emits v3 only and the agent rejects other versions.
- Centralized tunnel output parsing and documented the temporary VS Code workarounds with removal criteria.
- Removed the duplicated gateway deployment guide and the sample VPN implementation. Private connectivity is now an explicit prerequisite.
- Retained the portal as the workspace control plane. It now provides a native browser terminal, which removes the portal-to-CLI dead end while keeping portal and CLI ownership boundaries explicit.
- Moved the Microsoft relay approval requirement into prerequisites, centralized tool versions, made Microsoft the consistent tunnel default, and replaced DynamoDB scans with owner and state GSI queries.
- Corrected the architecture diagram boundaries so MicroVM, Lambda, DynamoDB, Bedrock, and S3 are outside customer subnets.

Validation: 206 tests pass, both TypeScript builds pass, and CDK synth passes in the default region and `eu-west-1`. Image 4.0 was restarted and manually checked in the UI; only `sonnet` is shown, with no duplicate `Custom` model.

Ready for re-review.

scouturier:
Re-reviewed at `497fcb1`. All eight items from the earlier review are addressed. Good to merge.

**Verified locally:** `tsc` clean, 206 tests pass, `cdk synth` clean under cdk-nag in `us-east-1`, `eu-west-1`, `ap-southeast-2`, and `eu-central-1`.

Issue #273, opened 2026-07-30
[Bug] Admin keys not injected into ECS task — spend_limits API returns auth.denied


kayalvizhi-kandasamy:
**Summary**
When a customer configures the admin: block in `gateway.yaml `with `write_keys `and `read_keys` (as documented), and invokes the `/v1/organizations/spend_limits` API with the `x-api-key` header, the gateway returns auth.denied with reason missing_token. The root cause could be that neither `setup.sh` nor `stamp-config.sh` nor the CDK stack provisions `GATEWAY_ADMIN_WRITE_KEY / GATEWAY_ADMIN_READ_KEY` as ECS task definition environment variables or Secrets Manager secrets.

Tested it with ECS Fargate deployment

**Steps to Reproduce**
1. Uncomment the admin: block in gateway.yaml.template:

```
admin:
  write_keys:
    - { id: terraform, key: "${GATEWAY_ADMIN_WRITE_KEY}" }
  read_keys:
    - { id: reporting, key: "${GATEWAY_ADMIN_READ_KEY}" }
  admin_groups: [platform-finops]
```

2. Deploy the gateway using setup.sh (or cdk deploy)
3. Invoke the spend_limits API as documented:

```
curl -sS https://<gateway-url>/v1/organizations/spend_limits/effective \
  -H "x-api-key: <the-key-value-you-intended>"
```

**Expected result:** Successful authentication and response from the spend_limits API.

**Actual result:**

```
{
  "ts": "2026-07-29T10:59:40.953Z",
  "evt": "auth.denied",
  "request_id": "023a4f81-1f8c-4694-937f-47e7a873e331",
  "reason": "missing_token",
  "path": "/v1/organizations/spend_limits/effective",
  "client_ip": "[IP_ADDRESS]"
}

```

**Documentation Gap:** Documentation needs to be updated with including the below.

- How to provision GATEWAY_ADMIN_WRITE_KEY / GATEWAY_ADMIN_READ_KEY in the selected deployment
- Any manual steps required beyond uncommenting the YAML block
- Keys management - additions/ deletion

bluedoors:
`gateway.yaml` is **baked into the image**, not mounted on ECS a task definition's
`secrets:` injects env vars only, and the image is distroless with no shell to
template at boot. So editing the config has no effect until a new image is built *and* the
service is pointed at it.

[...]

PR #274, opened 2026-07-31
fix(claude-apps-gateway): key the image tag on the baked config; document admin-API enablement

bluedoors:
Fixes the deployment trap behind #273.

## The problem

`gateway.yaml` is baked into the image, not mounted (on ECS a task definition's
`secrets:` injects env vars only, and the image is distroless with no shell to template
at boot). But the image tag was derived from `CLAUDE_VERSION` alone, while the image is
really *binary + config*.

So a config-only change produced no new tag:

- **`setup.sh`** — the idempotency check skipped the build (`image tag ... already in
  ECR`) while still re-registering the task definition and rolling the service. The
  deploy looks successful; the new tasks pull the same old image.
- **CDK (`docs/deployment.md`)** — same root issue, quieter: the documented push uses the
  bare version tag, so a rebuild overwrites it in place, and re-running pass 2 with an
  unchanged `-c imageTag` leaves the task definition identical, so ECS may not redeploy
  at all.

In #273 that surfaced as an admin-API failure: uncommenting the `admin:` block and
redeploying left the container running its previous config, so the admin routes were
never mounted and the request fell through to developer-session auth — returning
`auth.denied` / `missing_token` rather than `admin.denied` / `invalid_key`. The reported
symptom was therefore *not* caused by the missing admin-key provisioning (that would
crash-loop the container on `undefined env var in config`, not serve a 401).

## Changes

- **`cdk/scripts/setup.sh`** — new `config_image_tag()` helper; the image tag now
  defaults to `<version>-<config-hash>`, so any template edit yields a new tag and a
  genuine rebuild. Config stamping moved ahead of the ECR check (the tag depends on the
  stamped config), and the skip message now means "same binary *and* config".
  `IMAGE_TAG` still overrides to pin an explicit tag.
- **`docs/deployment.md`** — Track B never said a config change needs a new image tag.
  Documented, with the config-hash recipe.
- **`README.md` §5** — replaced the prose "inject them into the container" step with
  working CDK and `setup.sh` code; corrected task role → **execution** role (the ECS
  agent resolves secrets before the container starts); pointed step 3 at
  `cdk/gateway.yaml.template` rather than the generated `gateway.yaml`, which
  `stamp-config.sh` overwrites; and called out that all three steps must land in the
  same deploy, since half-applying fails two different ways and only one looks like an
  auth error.
- **`cdk/gateway.yaml.template` / `.example`** — documented `admin_groups` and
  `blocked_message` (the template and README each showed only one), noted the block is
  validated strictly so unknown keys fail the boot, and warned that uncommenting alone
  crash-loops the container.
- **`test/setup-helpers.test.sh`** — five cases covering the tag derivation: config
  change yields a new tag, determinism across re-runs, version-bump sensitivity, and
  ECR tag validity.

No new configuration surface or feature flags — the fix is a corrected default plus
documentation, keeping this a worked example rather than a product.

## Verification

Offline (no AWS account needed): `bash -n`, `shellcheck`, `test/setup-helpers.test.sh`
(19 passed), `test/stamp-config.test.sh` (9 passed), `cd cdk && npm test` (17 passed),
`npx cdk synth`. The uncommented `admin:` block was also parsed and checked against the
gateway's strict admin schema.

Live, on an ECS Fargate deployment (us-east-1):

1. **Reproduced #273** — a valid-length `x-api-key` against a gateway with no `admin:`
   block logged exactly `{"evt":"auth.denied","reason":"missing_token","path":"/v1/organizations/spend_limits/effective"}`,
   with the container healthy (2/2, `/healthz` 200) — confirming a stale image rather
   than a boot failure.
2. **Enabled admin correctly** — the same wrong-key request then returned
   `admin.denied` / `invalid_key`, proving the router mounts. Correct write key →
   `200`; `POST` a cap → `200`; the same `POST` with the *read* key → `403 requires
   write:spend_limits`.
3. **Demonstrated the tag fix on real artifacts** — same binary, two stamped configs:
   the old scheme produced `2.1.218` for both; the new scheme produced
   `2.1.218-3cb07511f94c` and `2.1.218-53c52556eb1e`.
4. **Full teardown + fresh CDK deploy** following the rewritten docs verbatim — stack
   `UPDATE_COMPLETE`, 2/2 tasks, both ALB targets healthy, image tag
   `2.1.218-39e683173ebe` from the documented recipe; `dig` → private IPs only,
   `/healthz` 200, `/readyz` `ready`, discovery doc valid, `device_authorization`
   issuing codes.

Two README claims were also confirmed against the live deployment: the execution role
(not the task role) is what holds `secretsmanager:GetSecretValue`, and
`openssl rand -base64 32` yields 44 characters, clearing the schema's 32-character
minimum.

Refs #273

### DIFF
--- a/claude-apps-gateway/README.md
+++ b/claude-apps-gateway/README.md
@@ -303,10 +303,21 @@ counters, an audit log, and per-developer PII. The example's RDS is deliberately
 disposable (`removalPolicy: DESTROY`, `deletionProtection: false`, 1-day backups,
 single-AZ), so turning this on is a conscious opt-in, not a default.
 
-**How to enable it** — three steps:
+**How to enable it** — three steps. This is a deliberate opt-in, so neither track
+provisions the admin keys for you; the code you need is below.
 
-**1. Mint the admin keys as secrets** (≥32 chars each), the same way the JWT/OIDC
-secrets are created:
+> [!IMPORTANT]
+> All three steps must land **together**, in the same deploy. The gateway expands
+> `${GATEWAY_ADMIN_WRITE_KEY}` when it reads the config, so a half-applied change
+> fails in one of two ways:
+> - **Config block without the secrets injected** → the container dies at boot with
+>   `undefined env var in config: GATEWAY_ADMIN_WRITE_KEY` (a crash-loop, not a 401).
+> - **Secrets injected without the config block** → the admin routes are never
+>   mounted, and requests fall through to developer-session auth, so `curl` gets
+>   `auth.denied` / `missing_token` even though the key is correct.
+
+**1. Mint the admin keys as secrets** (≥32 chars each — `openssl rand -base64 32`
+gives 44), the same way the JWT/OIDC secrets are created:
 
 ```bash
 aws secretsmanager create-secret --name claude-gateway-admin-write-key \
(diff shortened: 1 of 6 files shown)


bsnehanshu (approved):
traced `config_image_tag`'s version+config-hash logic through `setup.sh`'s phase-2 reorder — `IMAGE_URI`/`IMAGE_TAG` are only used after assembly, no ordering bug. `test/setup-helpers.test.sh` isn't wired into CI (`build.yml` only runs a repo-wide `yarn build`), so I ran it directly against this branch: 19/19 pass, including the 5 new `config_image_tag` cases. README's execution-role claim that ECS `secrets:` resolves via the task execution role matches [AWS ECS docs](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/specifying-sensitive-data-tutorial.html). Single scoped commit.

PR #270, opened 2026-07-27
docs(claude-apps-gateway): fix geo cross-region inference prefix — au, not apac

bsnehanshu:
## What

Every reference to the `apac.` geo cross-region inference prefix in `claude-apps-gateway`
was wrong. Anthropic's models on Bedrock use `au.` for the Australia/APAC geo profile, not
`apac.` — there is no `apac.` prefix. Additionally, `jp.` exists as a fourth geo prefix for
some models (Opus 4.8, Haiku 4.5) but not others (Sonnet 5 has no `jp.` profile).

Fixed in:
- `CLAUDE.md` — the "Bedrock IAM needs two ARN families" constraint note
- `README.md` — the IAM policy section + the `gateway.yaml` inline comment
- `cdk/README.md` — the "Regions & data residency" geo-prefix instructions
- `docs/deployment.md` — the model-ID resolution troubleshooting note
- `cdk/gateway.yaml.template` — the model-catalog inline comment
- `cdk/gateway.yaml.example` — the model-catalog inline comment
- `cdk/test/gateway-stack.test.ts` — a comment referencing the same (incorrect) prefix list

Also tracks `cdk/package-lock.json` (generated by `npm install` for this change), for the
same reason as #265 — reproducible synth/test and Dependabot coverage of transitive deps.

## Verification

- Checked the actual Bedrock model cards (`model-card-anthropic-claude-opus-4-8.html`,
  `model-card-anthropic-claude-sonnet-5.html`) via the AWS Documentation MCP server —
  Opus 4.8's Geo inference IDs are `us.`, `eu.`, `jp.`, `au.anthropic.claude-opus-4-8`;
  Sonnet 5's are `us.`, `eu.`, `au.anthropic.claude-sonnet-5` (no `jp.`). No `apac.` prefix
  exists on either.
- Repo-wide `grep -rn "apac\."` (no file-type filter) — zero matches, confirmed clean.
  An earlier pass only checked `.md`/`.yml`/`.ts` files and missed the two
  `gateway.yaml.template`/`.example` occurrences; this pass is unfiltered.
- `python3 -c "import yaml; yaml.safe_load(open('gateway.yaml.example'))"` — valid YAML
  (`.template` is not YAML-parseable as-is — it has `@@PLACEHOLDER@@` tokens by design,
  same as before this change; not a regression)
- `cd cdk && npm test` — 17/17 passed, including the test whose comment changed
- `cd cdk && npx cdk synth -c imageReady=false` — exit 0, clean synth

## Note

[...]

PR #277, opened 2026-08-04
docs(claude-apps-gateway): document the Claude Desktop overlay + the parent-settings opt-in

bluedoors:
## What

Anthropic added official Claude Desktop support to the apps gateway. This documents it. Two
separate mechanisms were previously missing or wrong here — one of them (2) is a silent
correctness bug for anyone who followed the existing snippets.

**1. Desktop as a gateway client.** The client half is `bootstrapUrl` →
`<public_url>/user/bootstrap` in Desktop's *own* managed configuration — a different key from
the CLI's `forceLoginGatewayUrl`. The server half is an explicit opt-in: `/user/bootstrap`
returns `404` unless the matching policy carries a `desktop` key, so `desktop: {}` on the
`match: {}` base opts in everyone who inherits it. Requires gateway ≥ 2.1.203 (this example
pins 2.1.218).

**2. Policy delivery to Desktop's embedded Claude Code sessions.** All three
`managed-settings.json` snippets were missing `parentSettingsBehavior: "merge"`. Claude Code
ignores parent settings on a machine with an admin-deployed managed source unless the winning
source sets it — and Desktop delivers gateway policy to the sessions it launches *as* parent
settings. Following the old snippet left those sessions **unpoliced, with no warning**. This
is the correctness fix in here.

The FAQ answer was also wrong ("configure the gateway URL in the desktop app's managed
settings") — it names the wrong key and omits the server-side opt-in.

## Two traps found while validating against a live deployment

- **`banner: { text: … }` renders nothing.** Desktop's `banner.enabled` has no default and
  `text` only applies when it's truthy, so the doc-style example needed `enabled: true`.
- **The `desktop` block is validated strictly and its key set is bounded by the pin** — an
  unknown key fails boot and crash-loops the ECS task. Verified against 2.1.218–2.1.221: nine
  keys accepted, and `chatTabEnabled` is *not* among them, so a bootstrap-configured Desktop
  loses its Chat tab with no way to restore it. Filed upstream as
  [anthropics/claude-code#83723](https://github.com/anthropics/claude-code/issues/83723) and
  noted in `docs/upstream-watch.md` with a probe recipe for future bumps.

## Scope

Docs + config template only. `/user/bootstrap` is same-host/same-port, so the ALB needs no new
listener rule or target group and **the CDK stack is unchanged**. `gateway.yaml.template` and
`gateway.yaml.example` kept in sync.

## Verification

Existing suites pass unchanged, no new tests (no code paths added):

- `cd cdk && npm test` — 17 passed
- `./test/stamp-config.test.sh` — 9 passed
- `./test/setup-helpers.test.sh` — 14 passed
- `bash -n` clean on all four `cdk/scripts/*.sh`
- `yaml.safe_load` clean on `gateway.yaml.example`

Capability 1 (`bootstrapUrl` → bootstrap config served, and the `404` when the opt-in is
missing) and both traps were confirmed against a live deployment; that's where the
`chatTabEnabled` gap and the `banner.enabled` behavior came from.

Rebased onto current `main`; no overlap with the `au.`/`jp.` prefix fix in #270.

### DIFF
--- a/claude-apps-gateway/README.md
+++ b/claude-apps-gateway/README.md
@@ -214,7 +214,7 @@ managed:
       desktop:                    # presence of this key is the opt-in
         isLocalDevMcpEnabled: false
         disableAutoUpdates: true
-        banner: { text: "Contractor build: internal use only" }
+        banner: { enabled: true, text: "Contractor build: internal use only" }
 ```
 
 The gateway derives most of the bootstrap response from the matched policy's `cli` block:
--- a/claude-apps-gateway/README.md
+++ b/claude-apps-gateway/README.md
@@ -109,7 +109,7 @@ Both the gateway server (Linux binary) and each developer's Claude Code CLI must
 
 Developers can update with `claude update`. The gateway server uses the same binary, downloaded from the Claude Code release page and packaged into a container image.
 
-Some gateway behaviour is version-gated: v2.1.198 added cross-upstream failover on `404` and the `anthropicAws` (Claude Platform on AWS) provider — earlier gateway builds reject that provider at boot. The worked example in this repo pins **2.1.218**, which also fixes gateway spend metering to price Bedrock application-inference-profile ARNs and other config-mapped upstream model IDs at the configured model's rates — directly relevant to this example's `global.anthropic.*` inference profiles. See [`docs/upstream-watch.md`](docs/upstream-watch.md) for a checklist to stay across gateway releases.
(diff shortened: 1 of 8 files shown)


bsnehanshu (approved):
Approved: Desktop-overlay claims (`bootstrapUrl`/`/user/bootstrap` opt-in, `desktop:` key set, audit event names, `parentSettingsBehavior: "merge"` precedence) verified verbatim against the live [config reference](https://code.claude.com/docs/en/claude-apps-gateway-config#claude-desktop-overlay) — including the `banner.enabled` gap, which Anthropic's own doc example has too. Independently reproduced the `chatTabEnabled` rejection from [anthropics/claude-code#83723](https://github.com/anthropics/claude-code/issues/83723) locally (schema rejects the key, passes once removed) — also confirmed on 2.1.207, below the issue's tested range. Ran all three claimed suites on this branch — `cdk test` 17/17, `setup-helpers.test.sh` 14/14, `stamp-config.test.sh` 9/9 — plus `bash -n` on all four scripts and YAML parse on `gateway.yaml.example`, all clean. Docs + config template only, no code paths added.

PR #284, opened 2026-08-19
feat: add claude-code-on-agentcore-runtime-microvm sample

wirjo:
## Summary

This sample gives developers an on-demand, private coding sandbox: each session provisions an isolated microVM (Claude Code CLI + VS Code CLI pre-installed) behind an IAM-authorized API, with an interactive shell and S3-backed checkpoint/restore so a workspace survives across sessions. It's a sibling to the existing `claude-code-on-lambda-microvm` sample — same governed developer-sandbox architecture and API shape, but running on **Amazon Bedrock AgentCore Runtime** (microVM compute type) instead of Lambda MicroVMs. Purely additive — no existing sample or shared file is modified.

## Architecture

```mermaid
flowchart LR
    CLI["Operator CLI<br/>(client/)"]
    Portal["Browser Portal<br/>optional, Cognito-gated"]

    subgraph AWS["AWS account"]
        APIGW["API Gateway<br/>IAM-authorized"]
        CP["Control Plane<br/>session lifecycle service"]
        DDB[("DynamoDB<br/>sessions / claims")]
        S3[("S3<br/>checkpoint bucket")]
        AR["AgentCore Runtime<br/>microVM container<br/>Claude Code CLI + VS Code CLI"]
    end

    CLI -- "REST (SigV4)<br/>bootstrap / terminate" --> APIGW --> CP
    CLI -- "WSS (SigV4)<br/>InvokeAgentRuntimeCommandShell" --> AR
    Portal -.-> AR
    CP -- "bootstrap / invoke / checkpoint" --> AR
    CP <--> DDB
    AR -- "checkpoint.tar.gz on terminate" --> S3
    AR -- "restore on bootstrap" --> S3
```

## Why AgentCore Runtime instead of Lambda MicroVM

AgentCore Runtime offers a managed, VPC-capable microVM compute surface with an equivalent 8h session cap, interactive shell access (`InvokeAgentRuntimeCommandShell`), and container-based deployment — but with a materially different API surface (no native session list/describe API, no pause primitive) that this sample works around explicitly rather than papering over.

## What's included

- **`infra/`** — CDK stack using the native `AWS::BedrockAgentCore::Runtime` L1 construct (`aws-cdk-lib` 2.260.0): VPC, KMS, S3 checkpoint bucket, DynamoDB sessions/claims tables, private IAM-authorized API Gateway, optional Cognito-backed browser portal (`enablePortal` context flag). `cdk-nag` (AwsSolutions) clean with documented suppressions.
- **`control-plane/`** — session lifecycle service on the AgentCore Runtime data-plane SDK. Emulates suspend/resume (no native AgentCore primitive) via container-side checkpoint/restore to S3, since AgentCore Runtime has no first-class session-describe or list API (`ListSessions` in the SDK is actually an AgentCore *Memory* API, not Runtime).
- **`agent-runtime/`** — container image (Claude Code CLI + VS Code CLI) implementing the AgentCore Runtime HTTP contract (`GET /ping`, `POST /invocations`, port 8080) for bootstrap/checkpoint/terminate.
- **`client/`** — operator CLI + a hand-rolled SigV4-signed WebSocket client for the interactive-shell protocol, since the JS AWS SDK doesn't expose `InvokeAgentRuntimeCommandShell` directly (confirmed against the `bedrock_agentcore` Python SDK source for the real wire protocol: `wss://bedrock-agentcore.<region>.amazonaws.com/runtimes/<arn>/ws/shells?shellId=<id>`, session id as a signed header, not a query param).
- **`scripts/`** — `provision-agent-image.ts` (builds/pushes the container to ECR, which is deliberately *not* CDK-owned since `CfnRuntime` needs an existing image at stack-create time), `deploy.ts`, `smoke-test.ts`.
- **`portal/`** — optional minimal browser terminal UI, mirroring the Lambda MicroVM sample's portal pattern (opt-in via `enablePortal`).

## Validation

Deployed and torn down against a real AWS account (account/region/profile kept generic in all shipped code and docs — see `deployment.example.json`):

- `npm run build` (tsc --noEmit) clean.
- `npm test` — 36/36 passing (control-plane service unit tests + CDK `Template` assertions).
- `cdk synth` clean, zero `cdk-nag` (AwsSolutions) findings.
- Real end-to-end deployment: `cdk deploy` succeeded, `AWS::BedrockAgentCore::Runtime` reached `READY`.
- Real interactive shell exercised via `InvokeAgentRuntimeCommandShell`:
  ```
  root@localhost:/# echo hello-from-agentcore-runtime
  hello-from-agentcore-runtime
  root@localhost:/# python3 --version
  Python 3.12.14
  root@localhost:/# whoami
  root
  ```
- Real checkpoint flow: bootstrap session (run-hook payload with S3 presigned upload URL) → wrote a file under `/workspace` via shell → triggered `terminate` via `InvokeAgentRuntime` → confirmed the checkpoint `.tar.gz` landed in the S3 workspace bucket with the correct size:
  ```
  1. bootstrap
  {"response":{"status":"initialized","sessionId":"..."},"status":"success"}
  2. checkpoint (terminate)
  {"response":{"status":"checkpointed","operation":"terminate"},"status":"success"}
  3. checking S3
  CHECKPOINT FOUND: s3://.../checkpoint.tar.gz size=355 bytes
  ```
- Stack torn down afterward (`cdk destroy`); no billable resources left running. (A few AWS-managed ENIs for the AgentCore Runtime persisted past deletion per AWS's documented up-to-8h behavior — handled via `--retain-resources` on the otherwise-empty VPC/subnet/security-group, no cost impact.)

## Notes for reviewers

- Deliberately does **not** modify `claude-code-on-lambda-microvm/` or any shared root config beyond the root `.gitignore`, where I added the same `!claude-code-on-agentcore-runtime-microvm/**/{bin,lib,test}/**` allowlist pattern that already exists for `claude-code-on-lambda-microvm/` (the repo's blanket `**/bin/`, `**/lib/`, `**/test/` ignore rules would otherwise silently drop this sample's CDK infra and test files).
- A comparison table against `claude-code-on-lambda-microvm` and the AgentCore-Runtime-specific tradeoffs (no native pause primitive, no session list/describe API, 1h shell-connection TTL requiring client-side reconnect) are documented in code comments throughout `infra/lib/platform-stack.ts` and `control-plane/src/aws-adapters.ts`; happy to pull these into a top-level README/deployment-guide if that's preferred before merge.

### DIFF
--- a/claude-code-on-agentcore-runtime-microvm/agent-runtime/agent.py
+++ b/claude-code-on-agentcore-runtime-microvm/agent-runtime/agent.py
@@ -1,0 +1,668 @@
+#!/usr/bin/env python3
+"""AgentCore Runtime HTTP protocol contract server and shell/session helpers.
+
+Implements GET /ping and POST /invocations per the AgentCore Runtime HTTP
+protocol contract (port 8080, host 0.0.0.0). Also runs the Claude Code
+workspace lifecycle: restore /workspace from the S3 checkpoint on the first
+invocation for a runtimeSessionId, and checkpoint back to S3 on a "suspend"
+or "terminate" lifecycle command sent through /invocations.
+
+Terminal access itself goes through AWS's InvokeAgentRuntimeCommandShell,
+which spawns its own PTY session directly against this container (see
+docs/deployment-guide.md); this script does not implement the shell
+WebSocket protocol -- AWS's own service handles that framing. This script
+only handles the one-shot /invocations lifecycle commands the control plane
+sends (session bootstrap, suspend, terminate) and workspace checkpoint I/O.
+"""
+
+from __future__ import annotations
+
+import base64
+import binascii
+import gzip
+import hashlib
+import http.client
+import io
+import json
+import logging
+import os
+import pwd
+import re
+import shutil
+import ssl
+import subprocess
+import sys
+import tarfile
+import tempfile
+import threading
+import time
+from dataclasses import dataclass
+from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
+from pathlib import Path, PurePosixPath
+from typing import Any
+from urllib.parse import parse_qs, urlsplit
+
+LOG = logging.getLogger("claude-agentcore-agent")
+
+WORKSPACE = Path("/workspace")
+DEVELOPER_HOME = WORKSPACE / ".claude-home"
+CLAUDE_SETTINGS = DEVELOPER_HOME / ".claude" / "settings.json"
+STATE_DIRECTORY = Path("/var/lib/claude-agentcore")
+SESSION_CONFIGURATION = STATE_DIRECTORY / "session.json"
+CLAUDE_BINARY = Path("/usr/local/bin/claude")
+
+WORKSPACE_ID_PATTERN = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]{0,63}$")
+SESSION_ID_PATTERN = re.compile(r"^[A-Za-z0-9-]{1,128}$")
+OWNER_HASH_PATTERN = re.compile(r"^[a-f0-9]{64}$")
+AWS_REGION_PATTERN = re.compile(r"^[a-z]{2}(?:-gov)?-[a-z]+-\d$")
+BEDROCK_MODEL_PATTERN = re.compile(
+    r"^(?:(?:us|eu|au|global)\.)?"
+    r"anthropic\.claude-[A-Za-z0-9._:-]{1,180}$"
+)
+
+MAX_RUN_HOOK_PAYLOAD_BYTES = 4 * 1024
+MAX_DECODED_RUN_HOOK_PAYLOAD_BYTES = 16 * 1024
+COMPRESSED_RUN_HOOK_PAYLOAD_PREFIX = "gzip-base64:"
+MAX_CHECKPOINT_BYTES = int(
+    os.environ.get("MAX_CHECKPOINT_BYTES", str(128 * 1024 * 1024))
+)
+MAX_EXTRACTED_BYTES = int(
+    os.environ.get("MAX_EXTRACTED_BYTES", str(1024 * 1024 * 1024))
+)
+MAX_ARCHIVE_MEMBERS = 200_000
+CHECKPOINT_TIMEOUT_SECONDS = 50
+
+
+@dataclass(frozen=True)
+class Session:
+    session_id: str
+    owner_hash: str
+    workspace_id: str
+    aws_region: str
+    inference_mode: str
+    bedrock_model_id: str | None
+    checkpoint_download_url: str | None
+    checkpoint_upload_url: str
+    access_mode: str = "terminal"
+
+
+class Runtime:
+    def __init__(self) -> None:
+        self._lock = threading.RLock()
+        self._session: Session | None = None
+        self._checkpoint_current = False
+
+    def bootstrap(self, payload: str) -> dict[str, Any]:
+        session = parse_run_hook_payload(payload)
+        with self._lock:
+            if self._session and self._session != session:
+                raise ValueError("A different session is already initialized")
+            if self._session:
+                return {"status": "already-initialized"}
+
+            STATE_DIRECTORY.mkdir(parents=True, exist_ok=True, mode=0o700)
+            checkpoint = download_checkpoint(session)
+            try:
+                if checkpoint:
+                    restore_workspace(checkpoint)
+                else:
+                    reset_workspace()
+            finally:
+                if checkpoint:
+                    checkpoint.unlink(missing_ok=True)
+
+            initialize_developer_home()
+            write_terminal_defaults()
+            write_session_configuration(session)
+            self._session = session
+            self._checkpoint_current = False
+            LOG.info(
+                "session initialized",
+                extra={"session_id": session.session_id},
+            )
+        return {"status": "initialized", "sessionId": session.session_id}
+
+    def checkpoint(self, operation: str) -> dict[str, Any]:
+        with self._lock:
+            if not self._session:
+                return {"status": "no-active-session"}
+            if self._checkpoint_current:
+                return {"status": "already-current"}
+            archive = create_workspace_archive()
+            try:
+                upload_checkpoint(self._session, archive)
+                self._checkpoint_current = True
+            finally:
+                archive.unlink(missing_ok=True)
+        LOG.info("workspace checkpoint uploaded", extra={"operation": operation})
+        return {"status": "checkpointed", "operation": operation}
+
+
+RUNTIME = Runtime()
+
+
+def parse_run_hook_payload(value: str) -> Session:
+    decoded = decode_run_hook_payload(value)
+    try:
+        payload = json.loads(decoded)
+    except json.JSONDecodeError as error:
+        raise ValueError("runHookPayload is not valid JSON") from error
+    if not isinstance(payload, dict) or payload.get("version") != 1:
+        raise ValueError("Unsupported run hook payload")
+
+    session_id = required_string(payload, "sessionId", 128)
+    owner_hash = required_string(payload, "ownerHash", 64)
+    workspace_id = required_string(payload, "workspaceId", 64)
+    aws_region = required_string(payload, "awsRegion", 32)
+    if not AWS_REGION_PATTERN.fullmatch(aws_region):
+        raise ValueError("Invalid awsRegion")
+    inference_mode = required_string(payload, "inferenceMode", 32)
+    if inference_mode not in ("bedrock", "claude-ai", "claude-gateway"):
+        raise ValueError("Unsupported inferenceMode")
+    bedrock_model_id = (
+        required_string(payload, "bedrockModelId", 220)
+        if inference_mode == "bedrock"
+        else None
+    )
+    if bedrock_model_id is not None and not BEDROCK_MODEL_PATTERN.fullmatch(
+        bedrock_model_id
+    ):
+        raise ValueError("Invalid bedrockModelId")
+    checkpoint = payload.get("checkpoint")
+    if not isinstance(checkpoint, dict):
+        raise ValueError("checkpoint must be an object")
+    download_value = checkpoint.get("downloadUrl")
+    checkpoint_download_url = (
+        download_value if isinstance(download_value, str) and download_value else None
+    )
+    checkpoint_upload_url = required_string(checkpoint, "uploadUrl", 8_000)
+
+    if not SESSION_ID_PATTERN.fullmatch(session_id):
+        raise ValueError("Invalid sessionId")
+    if not OWNER_HASH_PATTERN.fullmatch(owner_hash):
+        raise ValueError("Invalid ownerHash")
+    if not WORKSPACE_ID_PATTERN.fullmatch(workspace_id):
+        raise ValueError("Invalid workspaceId")
+
+    return Session(
+        session_id=session_id,
+        owner_hash=owner_hash,
+        workspace_id=workspace_id,
+        aws_region=aws_region,
+        inference_mode=inference_mode,
+        bedrock_model_id=bedrock_model_id,
+        checkpoint_download_url=checkpoint_download_url,
+        checkpoint_upload_url=checkpoint_upload_url,
+    )
+
+
+def decode_run_hook_payload(value: str) -> str:
+    if not value.startswith(COMPRESSED_RUN_HOOK_PAYLOAD_PREFIX):
+        return value
+    encoded = value.removeprefix(COMPRESSED_RUN_HOOK_PAYLOAD_PREFIX)
+    try:
+        compressed = base64.b64decode(encoded, validate=True)
+        with gzip.GzipFile(fileobj=io.BytesIO(compressed)) as stream:
+            decoded = stream.read(MAX_DECODED_RUN_HOOK_PAYLOAD_BYTES + 1)
+    except (binascii.Error, EOFError, OSError) as error:
+        raise ValueError("runHookPayload compression is invalid") from error
+    if len(decoded) > MAX_DECODED_RUN_HOOK_PAYLOAD_BYTES:
+        raise ValueError("runHookPayload exceeds the decoded size limit")
+    return decoded.decode("utf-8")
+
+
+def required_string(value: dict[str, Any], key: str, max_length: int) -> str:
+    result = value.get(key)
+    if (
+        not isinstance(result, str)
+        or not result
+        or len(result.encode("utf-8")) > max_length
+    ):
+        raise ValueError(f"{key} must be a non-empty string")
+    return result
+
+
+def write_terminal_defaults() -> None:
+    developer = pwd.getpwnam("developer")
+    settings: dict[str, Any] = {}
+    if CLAUDE_SETTINGS.exists():
+        try:
+            value = json.loads(CLAUDE_SETTINGS.read_text(encoding="utf-8"))
+        except (OSError, json.JSONDecodeError):
+            value = {}
+        if isinstance(value, dict):
+            settings = value
+    settings.setdefault("theme", "dark-ansi")
+    settings.setdefault("tui", "fullscreen")
+    CLAUDE_SETTINGS.parent.mkdir(parents=True, exist_ok=True, mode=0o700)
+    os.chown(CLAUDE_SETTINGS.parent, developer.pw_uid, developer.pw_gid)
+    atomic_json_write(
+        CLAUDE_SETTINGS,
+        settings,
+        mode=0o600,
+        owner=developer.pw_uid,
+        group=developer.pw_gid,
+    )
+
+
+def session_environment(session: Session) -> dict[str, str]:
+    environment = {
+        "AWS_REGION": session.aws_region,
+        "AWS_DEFAULT_REGION": session.aws_region,
+        "HOME": str(WORKSPACE / ".home"),
+        "USER": "developer",
+        "SHELL": "/bin/bash",
+        "TERM": "xterm-256color",
+        "PATH": "/usr/local/bin:/usr/bin:/bin",
+        "CLAUDE_CONFIG_DIR": str(CLAUDE_SETTINGS.parent),
+        "DISABLE_AUTOUPDATER": "1",
+        "CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC": "1",
+    }
+    if session.inference_mode == "bedrock" and session.bedrock_model_id:
+        environment["CLAUDE_CODE_USE_BEDROCK"] = "1"
+        environment["ANTHROPIC_MODEL"] = session.bedrock_model_id
+    return environment
+
+
+def write_session_configuration(session: Session) -> None:
+    environment = session_environment(session)
+    developer = pwd.getpwnam("developer")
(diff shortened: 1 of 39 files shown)


claude-code-on-agentcore-runtime-microvm/scripts/deploy.ts
```diff
@@ -0,0 +1,235 @@
...
+  }
+  return configuration;
+}
+
+function platformParameterArguments(
+  configuration: DeploymentConfiguration,
+  projectName: string,
+): string[] {
+  const parameters: Record<string, string | number | undefined> = {
+    ProjectName: projectName,
```
scouturier, line 150: **Blocking: deploy fails on an undeclared parameter.**

The stack only declares two `CfnParameter`s (`IdleAfterSeconds` at `platform-stack.ts:81` and `TrustedClientCidr` at `:94`). Passing `--parameters ClaudeAgentCoreRuntimeStack:ProjectName=...` makes CloudFormation reject the deploy with an unknown-parameter error, so `npm run deploy` cannot succeed as committed.

The stack reads `projectName` from CDK context instead, which `deploy.ts` never sets. If the parameter is simply dropped, a custom `projectName` in `deployment.json` still synthesizes with the default `claude-agentcore`, so `CfnRuntime.containerUri` points at a repository that `provision-agent-image.ts` never pushed to.

Related: `allowClaudeAiSubscription` is passed as context but `platform-stack.ts` never reads it or exports `ALLOW_CLAUDE_AI_SUBSCRIPTION`, so setting it to `true` still yields the 403 at `service.ts:96-101`.

Suggest passing `projectName` and `allowClaudeAiSubscription` via `--context` (or declaring them as parameters) and adding a `Template` assertion that covers the parameter set `deploy.ts` sends.

claude-code-on-agentcore-runtime-microvm/control-plane/src/aws-adapters.ts
```diff
@@ -0,0 +1,514 @@
...
+    }
+    const startedAt = Math.floor(Date.now() / 1_000);
+    await this.client.send(
+      new InvokeAgentRuntimeCommandCommand({
+        agentRuntimeArn: input.agentRuntimeArn,
+        runtimeSessionId: input.runtimeSessionId,
+        contentType: 'application/json',
+        accept: 'application/json',
+        body: {
+          command: `/opt/claude-agentcore/session-bootstrap.sh '${input.payload}'`,
```
scouturier: **Blocking: bootstrap runs a script that does not exist in the image.**

`agent-runtime/Dockerfile` only copies `tool-versions.json` and `agent.py` into `/opt/claude-agentcore/`; there is no `session-bootstrap.sh`. `InvokeAgentRuntimeCommand` returns exit 127 in-band via `contentStop.exitCode` rather than throwing, and this code discards the exit code, so the record still flips to `RUNNING`.

Nothing else ever POSTs `{"command":"bootstrap", ...}` to `agent.py`'s `/invocations` either, so `Runtime.bootstrap()` never runs: `/workspace` is never restored from S3, `/var/lib/claude-agentcore/session.json` is never written, and the developer shell fails with `Session configuration is unavailable` (`agent.py:286`) on every session.

Either ship the script in the image, or (probably simpler) call `InvokeAgentRuntime` with the bootstrap payload here, and treat a non-zero exit code as a failed start.

claude-code-on-agentcore-runtime-microvm/control-plane/src/aws-adapters.ts
```diff
@@ -0,0 +1,514 @@
...
+    agentRuntimeArn: string,
+    runtimeSessionId: string,
+  ): Promise<void> {
+    const description = await this.get(agentRuntimeArn, runtimeSessionId);
+    if (description.state === 'TERMINATED') {
+      throw new Error('Cannot resume a terminated AgentCore Runtime session');
+    }
+  }
+
+  public async terminate(
```
scouturier, line 426: **Blocking: checkpoint is unreachable, so terminate silently loses the workspace.**

`terminate()` calls `StopRuntimeSession` directly and `suspend()` is an empty no-op. No code path sends the `suspend` or `terminate` lifecycle command to `/invocations`, so `Runtime.checkpoint()` in `agent.py`, the only place the `checkpoint.tar.gz` upload happens, is never invoked. The README and CLI help both promise "Checkpoint and terminate".

Scenario: developer edits `/workspace`, runs `terminate <id>` (or `terminateBeforeExpiry` fires), the microVM is stopped with no PUT, and the next `start` for the same `workspaceId` gets `NotFound` from `HeadObject` and resets to an empty workspace.

Once this is wired, two latent issues in `agent.py` will surface: `_checkpoint_current` is never reset after the first upload (`agent.py:134`), and an absolute symlink anywhere under `/workspace` (e.g. a `.venv`) aborts the whole archive (`agent.py:555`). I left notes on both lines.

claude-code-on-agentcore-runtime-microvm/control-plane/src/handler.ts
```diff
@@ -0,0 +1,345 @@
...
+  if (apiId && stage) {
+    observedApiUrl = `https://${apiId}.execute-api.${region}.amazonaws.com/${stage}`;
+    if (configurationCache) {
+      configurationCache.value.controlApiUrl =
+        process.env.CONTROL_API_URL || observedApiUrl;
+    }
+  }
+}
+
+function assertRuntimeExecutionCaller(principal: string): void {
```
scouturier, line 208: **Security: any sandbox can mint presigned URLs for another owner's checkpoint.**

`POST /sessions/{sessionId}/checkpoint-urls` authorizes by IAM role name, which every microVM shares (they all run under the same execution role). The `body.runtimeSessionId` check does not add anything because `service.ts:175` sets `runtimeSessionId` equal to the session UUID that is already in the path.

Scenario: a process in sandbox A (execution-role credentials, `execute-api:Invoke` on this route, runtime security group allowed to the execute-api VPC endpoint) learns session B's UUID and calls `POST /sessions/<B>/checkpoint-urls` with `{"runtimeSessionId":"<B>"}`. It receives 10-hour GET and PUT URLs for `workspaces/<ownerHashB>/<ws>/checkpoint.tar.gz` and can read or overwrite B's workspace.

The cdk-nag suppression in `nag-acks.ts` states that the handler verifies ownership, which it does not. Options: pass a per-session random secret to the container at bootstrap and require it here, or scope the presigned URLs to a per-session IAM session (e.g. `SourceIdentity` / session tags) rather than the shared role.

claude-code-on-agentcore-runtime-microvm/portal/site.ts
```diff
@@ -0,0 +1,339 @@
...
+  el('terminal-dialog').showModal();
+  terminal = new window.Terminal({ convertEol: true });
+  terminal.open(el('terminal-screen'));
+  connectTerminal(session);
+}
+
+async function connectTerminal(session) {
+  try {
+    var connection = await api('POST', 'sessions/' + session.sessionId + '/connect', {});
+    var socket = new WebSocket(connection.shellUrl);
```
scouturier, line 457: **Blocking for the portal: unsigned WebSocket upgrade is rejected.**

`InvokeAgentRuntimeCommandShell` requires a SigV4-signed upgrade plus the `X-Amzn-Bedrock-AgentCore-Runtime-Session-Id` header (this is exactly what `client/src/terminal.ts` does). Browsers cannot set custom headers on a WebSocket upgrade, and `authToken` is always `''` here, so the connect to `wss://bedrock-agentcore.<region>.amazonaws.com/...` fails with 403 and the terminal stays blank after README step 4.

Even if a relay made the connection succeed, this code ignores the channel-prefix framing from `shell-protocol.ts`: keystrokes go out without the leading `0x00` STDIN byte, received frames are written to xterm including the channel byte, and no bootstrap or privilege-drop command is sent, so the shell would be root.

The portal probably needs a server-side relay (Lambda or the control plane) that signs the upgrade and speaks the framed protocol, or it should be dropped from this PR and left as a follow-up rather than shipped non-functional behind `enablePortal`.

claude-code-on-agentcore-runtime-microvm/client/src/cli.ts
```diff
@@ -0,0 +1,258 @@
...
+      const inferenceMode = takeClaudeProvider(args);
+      const workspaceId = args.shift();
+      assertNoArguments(args);
+      const result = await client.start(workspaceId, {
+        accessMode: 'terminal',
+        inferenceMode,
+      });
+      printStartResult(result);
+      if (!noConnect) {
+        await attachTerminal(await client.connect(result.session.sessionId));
```
scouturier: **Blocking: shell runs as root with no Bedrock environment.**

Both `attachTerminal` call sites (here and line 47) omit the `bootstrapCommand` argument, so `developerShellBootstrapCommand()` in `terminal.ts:442` (the `setpriv` drop to uid 1000 plus the session-environment loader) is never sent. The Dockerfile also has no `USER` directive.

Result: `npm run client -- start` lands the operator in the root bash that `InvokeAgentRuntimeCommandShell` spawns, with `HOME=/root` and none of `CLAUDE_CODE_USE_BEDROCK`, `ANTHROPIC_MODEL`, or `CLAUDE_CONFIG_DIR` set. Running `claude` prompts for a login instead of using Bedrock, and every session runs as root inside the container.

Pass `developerShellBootstrapCommand()` at both call sites (or make it the default inside `attachTerminal`).

claude-code-on-agentcore-runtime-microvm/control-plane/src/aws-adapters.ts
```diff
@@ -0,0 +1,514 @@
...
+      maximumDurationInSeconds: RUNTIME_SESSION_MAX_DURATION_SECONDS,
+    };
+  }
+
+  public async get(
+    agentRuntimeArn: string,
+    runtimeSessionId: string,
+  ): Promise<RuntimeSessionDescription> {
+    try {
+      await this.client.send(
```
scouturier, line 346: **Correctness: the liveness probe resurrects ended sessions and defeats the idle timeout.**

`get()` probes with a real `InvokeAgentRuntimeCommand` (`command: 'true'`) on the stored `runtimeSessionId`. But `run()` in this same class relies on that exact call provisioning a new microVM for an unknown session id, so the probe cannot distinguish "alive" from "gone". Per the AgentCore docs it also counts as activity and resets `idleRuntimeSessionTimeout`.

Scenario: a session idles out or is stopped by AWS, the 1-minute `IdleReaperSchedule` fires `refreshFromRuntime`, `get()` invokes `true`, a fresh empty microVM is provisioned and reported `RUNNING`. The record never transitions, the workspace claim is never released, `resume()` always succeeds, and connect lands in an un-bootstrapped container. Separately, abandoned sessions can never hit the 900s idle timeout and run the full 8h `maxLifetime`.

`lastActivityAt` and `idleAfterSeconds` are stored but never read in `service.ts`. Since the Runtime API has no describe call, the control plane probably has to own idle detection from its own `lastActivityAt` and stop sessions itself, rather than probing.

claude-code-on-agentcore-runtime-microvm/control-plane/src/service.ts
```diff
@@ -0,0 +1,636 @@
...
+        'SUSPENDED',
+        'RESUMING',
+        'FAILED',
+      ],
+    );
+    if (!claimed) {
+      return (await this.options.repository.get(sessionId)) ?? record;
+    }
+    const config = await this.options.loadConfiguration();
+    await this.options.agentRuntime.terminate(
```
scouturier, line 398: **Correctness: `TERMINATING` is a dead end if the stop call throws.**

The record is patched to `TERMINATING` before this call, and there is no `try/catch` or compensating update. If `StopRuntimeSession` throws (throttling, transient 5xx), the request returns 500 and nothing ever retries or releases the workspace claim: later `DELETE`s return early at line 361, `refreshFromRuntime` returns unchanged at line 546 while the runtime still answers, and `POST /sessions` for that workspace returns the stuck record with `created: false`. The user is locked out until the microVM dies on its own, up to 8h.

`terminateBeforeExpiry` has the same unguarded pattern. Suggest either patching to `TERMINATING` after a successful stop, or letting the reconciler retry stop for records stuck in `TERMINATING`.

claude-code-on-agentcore-runtime-microvm/control-plane/src/service.ts
```diff
@@ -0,0 +1,636 @@
...
+        runtimeSessionId: run.runtimeSessionId,
+        runtimeStartedAt: run.startedAt,
+        runtimeExpiresAt: run.startedAt + run.maximumDurationInSeconds,
+      };
+      const updated = await this.options.repository.patch(
+        sessionId,
+        patch,
+        ['PROVISIONING'],
+      );
+      if (!updated) {
```
scouturier, line 195: **Correctness: orphaned microVM when the post-run write fails.**

Mirror image of the terminate issue: `run()` has already provisioned a microVM by this point, so if this conditional update fails there is no compensating `terminate()` and a live, billable microVM is left with no record pointing at it. A `try/catch` that stops the runtime session before rethrowing would close the gap.

claude-code-on-agentcore-runtime-microvm/client/src/terminal.ts
```diff
@@ -0,0 +1,470 @@
...
+}
+
+async function signShellUpgrade(
+  url: URL,
+  runtimeSessionId: string,
+): Promise<{ headers: Record<string, string> }> {
+  // Matches the real interactive-shell WebSocket contract (verified against
+  // the bedrock-agentcore Python SDK's AgentCoreRuntimeClient.connect_shell):
+  // the runtime session id travels as a signed header, not a query param.
+  const credentials = await defaultProvider()();
```
scouturier: **Correctness: `--profile` is ignored for the shell upgrade.**

`signShellUpgrade` signs with `defaultProvider()` and no profile, while `ControlApiClient` (`api.ts:76`) honours `--profile`. The profile is never threaded into `attachTerminal`.

Scenario: `npm run client -- --profile dev start` with no usable default profile. `POST /sessions` and `/connect` succeed via `dev`, then the WebSocket upgrade is signed with default-chain credentials (missing, expired, or a different account) and fails with 403 or "Unable to locate credentials", while the session keeps running and billing.

`scripts/smoke-test.ts:116` already passes the profile to its own copy of this function, which is also a hint that the four copies of `signShellUpgrade` should be consolidated into `shell-protocol.ts` or a shared module.

claude-code-on-agentcore-runtime-microvm/client/src/terminal.ts
```diff
@@ -0,0 +1,470 @@
...
+
+  outputListener = (data): void => {
+    lastReadAt = Date.now();
+    const frame = decodeShellFrame(asBuffer(data));
+    if (frame.channel === ShellChannel.Stdout) {
+      terminalOutput.write(frame.payload);
+    } else if (frame.channel === ShellChannel.Stderr) {
+      terminalOutput.write(frame.payload);
+    } else if (frame.channel === ShellChannel.Status) {
+      const status = parseShellStatus(frame.payload);
```
scouturier, line 219: **Correctness: a malformed frame crashes the CLI with the terminal left in raw mode.**

Both `ws` `'message'` listeners (`outputListener` and `waitForSessionInitialization`) call `decodeShellFrame` (throws on a 0-byte frame), `parseShellStatus` (bare `JSON.parse`), and `asBuffer` with no `try/catch`, and there is no `uncaughtException` handler. A synchronous throw inside an `EventEmitter` callback exits Node before the `finally` blocks run, so the restore sequence at lines 143-149 never executes and the user is left with stdin in raw mode and the alternate screen active.

This is most likely to bite during the hourly reconnect. Wrap the frame handling in `try/catch` and route errors through the same path that restores the terminal.

scouturier (commented):
Thanks for the thorough write-up. I checked out the head commit: `tsc --noEmit` is clean, all 44 vitest tests pass, and cdk-nag is clean, so the scaffolding is in good shape.

However, tracing the shipped control plane and CLI end to end turned up several places where the pieces are not wired together, so the session flows described in the README do not work through `npm run deploy` and `npm run client` as committed. The validation in the description looks like it exercised the container and API directly rather than through this code path. Inline comments below, roughly most severe first:

**Blocking**
1. `deploy.ts` passes a `ProjectName` CloudFormation parameter the stack never declares, so `npm run deploy` fails.
2. Bootstrap executes `/opt/claude-agentcore/session-bootstrap.sh`, which the Dockerfile never installs, and nothing ever sends the `bootstrap` command to `/invocations`.
3. No code path sends `suspend` or `terminate` to `/invocations`, so the S3 checkpoint upload is unreachable and every terminate loses the workspace.
4. The browser portal opens an unsigned WebSocket, which the service rejects.
5. The CLI never passes the developer-shell bootstrap command, so sessions run as root with no Bedrock environment.

**Security**
6. The checkpoint-urls route authorizes by shared role name plus a session UUID that is already in the path, so one sandbox can mint presigned URLs for another owner's workspace. The cdk-nag suppression says ownership is verified, but it is not.

**Correctness**
7. The liveness probe provisions a new microVM for ended sessions and resets the idle timer.
8. `TERMINATING` is a dead end if `StopRuntimeSession` throws, and a live microVM is orphaned if the post-run DynamoDB write fails.
9. `--profile` is ignored when signing the shell upgrade.
10. A malformed shell frame crashes the CLI with the terminal left in raw mode.

Plus a few smaller items in `agent.py`. Once the lifecycle commands are wired through `/invocations` and the deploy parameters fixed, it would be worth re-running the validation strictly via `npm run deploy` and `npm run client -- start` / `terminate` / `start` against the same workspace to confirm restore actually round-trips.

Also +1 to your own suggestion of moving the comparison table and tradeoffs from code comments into the README.

claude-code-on-agentcore-runtime-microvm/agent-runtime/agent.py
```diff
@@ -0,0 +1,668 @@
...
+    def checkpoint(self, operation: str) -> dict[str, Any]:
+        with self._lock:
+            if not self._session:
+                return {"status": "no-active-session"}
+            if self._checkpoint_current:
+                return {"status": "already-current"}
+            archive = create_workspace_archive()
+            try:
+                upload_checkpoint(self._session, archive)
+                self._checkpoint_current = True
```
scouturier, line 134: Minor (latent until checkpoint is wired): `_checkpoint_current` is set here after the first upload but never reset when the workspace changes afterwards, so a second `suspend`/`terminate` in the same session would skip the upload and lose edits made since the first checkpoint.

claude-code-on-agentcore-runtime-microvm/agent-runtime/agent.py
```diff
@@ -0,0 +1,668 @@
...
+        if member.isreg():
+            total_bytes += member.size
+            if total_bytes > MAX_EXTRACTED_BYTES:
+                raise RuntimeError("Workspace exceeds the checkpoint limit")
+            return member
+        if member.isdir():
+            return member
+        if member.issym():
+            relative = safe_member_path(member.name)
+            validate_symlink(relative, member.linkname)
```
scouturier, line 556: Minor (latent until checkpoint is wired): `validate_symlink` raises on any absolute symlink target, which aborts the whole archive. A Python `.venv` or many Node toolchains create absolute symlinks under the workspace, so a normal dev workspace would fail to checkpoint. Consider skipping or rewriting such members instead of failing, and note that the handler maps this `ValueError` to HTTP 400 rather than 500, which will make the failure look like a bad request from the control plane.

PR #288, opened 2026-08-29
fix(claude-apps-gateway): make deploy.sh safe to re-run — status-aware pass 1 + destructive-change gate

### LINKED ISSUE #287: claude-apps-gateway: re-running deploy.sh tears down the full stack (pass 1 runs unconditionally)
**Sample:** `claude-apps-gateway` (CDK track)

## Bug

Re-running `cdk/scripts/deploy.sh` against an already-deployed stack destroys it.

Step 1 runs CDK pass 1 unconditionally:

```bash
echo "Step 1/5: Pass 1 — creating the ECR repository (CDK)..."
npx cdk deploy --require-approval never -c imageReady=false "${CDK_CTX[@]}"
```

With `-c imageReady=false` the stack synthesizes a template containing **only the ECR repository** (that is the point of the two-pass deploy). Deploying that template as an *update* over an existing full stack makes CloudFormation remove every other resource: VPC, RDS, ALB, ECS service, secrets, DNS record. The script's stated purpose ("Deploys CDK stack, builds container image, and starts the service") invites re-running it for image/config updates, which is exactly the destructive case.

## Steps to reproduce

1. `./scripts/deploy.sh` — completes, gateway healthy.
2. `./scripts/deploy.sh` again (e.g. to roll a new image).
3. Pass 1 deploys the repo-only template → CloudFormation deletes the VPC/RDS/ALB/service.

## Related failure modes found while fixing

- A failed **first** deploy leaves the stack in `ROLLBACK_COMPLETE`. A naive existence check then skips pass 1, `EcrRepositoryUri` reads back as `None`, and the CodeBuild IAM policy is created scoped to `repository/None` — the run fails far from the cause. (`ROLLBACK_COMPLETE` stacks can't be updated either; they must be deleted.)
- The CodeBuild wait loop only treats `SUCCEEDED|FAILED|STOPPED` as terminal; a `TIMED_OUT` or `FAULT` build spins it forever.

## Proposed fix

PR incoming (per CONTRIBUTING, opening this issue first):

- Switch step 1 on the stack **status**: healthy → skip pass 1; absent → run pass 1; `ROLLBACK_COMPLETE` → fail early with `delete-stack` recovery commands; in-progress → fail with wait-and-retry.
- Validate the `EcrRepositoryUri` output before use.
- Add `FAULT`/`TIMED_OUT` as terminal build states.
- Gate pass 2 on `cdk diff`: abort when the change set would remove or replace resources, with `FORCE_DESTRUCTIVE=1` as an explicit override for intentional changes.

Environment: verified on a live deployment (ap-southeast-7), aws-cdk 2.1133.0, macOS bash 3.2 client.

peepeepopapapeepeepo:
Fixes #287.

## Problem

`cdk/scripts/deploy.sh` ran CDK pass 1 (`-c imageReady=false`) unconditionally. Pass 1 synthesizes a template containing only the ECR repository, so re-running the script against a live stack updates it to that template — CloudFormation removes the VPC, RDS, ALB, ECS service, secrets, and DNS record. Details and repro in #287.

## Changes (all in `claude-apps-gateway/cdk/scripts/deploy.sh`)

1. **Status-aware step 1.** `describe-stacks` now feeds a `case` on `StackStatus` instead of running pass 1 blindly:
   - stack absent → run pass 1 (first deploy, unchanged behaviour)
   - `CREATE_COMPLETE` / `UPDATE_COMPLETE` / `UPDATE_ROLLBACK_COMPLETE` → skip pass 1, go straight to build + pass 2
   - `ROLLBACK_COMPLETE` (failed first deploy — not updatable) → fail early with the exact `delete-stack` + `wait` recovery commands
   - any in-progress state → fail with a wait-and-retry message
2. **`EcrRepositoryUri` output guard.** A partial/rolled-back stack returns `""`/`None`; previously that flowed into the CodeBuild IAM policy (`repository/None`) and failed far from the cause. Now it fails immediately with a pointer to the console.
3. **Terminal build states.** The CodeBuild poll treats `FAULT` and `TIMED_OUT` as failures; previously they spun the loop forever.
4. **Destructive-change gate on pass 2.** `cdk diff` runs first; the deploy aborts if the change set would remove (`^\[-\]`) or replace (`requires replacement` / `may be replaced`) any resource. `FORCE_DESTRUCTIVE=1` overrides for intentional changes (e.g. dropping VPC endpoints after `CREATE_VPC_ENDPOINTS=false`). The match is deliberately narrow — a bare `replace` grep would trip on cdk's own boilerplate line on every run. Synth errors are echoed before exiting instead of dying silently under `set -e` (`cdk diff` exits 0 on mere differences unless `--fail`; verified on 2.1133.0).

## Testing

- `bash -n` clean.
- `cdk` suite: `npm test` — 17/17 pass (no template changes; script-only).
- Branch logic exercised for `NONE`, all three healthy states, `ROLLBACK_COMPLETE`, and in-progress states.
- Live re-run against an `UPDATE_COMPLETE` deployment (ap-southeast-7) lands on the skip path and proceeds to build + pass 2.

By submitting this pull request, I confirm that my contribution is made under the terms of the repository's license.

### DIFF
--- a/claude-apps-gateway/cdk/scripts/deploy.sh
+++ b/claude-apps-gateway/cdk/scripts/deploy.sh
@@ -95,15 +95,48 @@ fi
 # The ECS service can't start until its image exists, so the stack splits the
 # deploy in two: pass 1 creates just the ECR repo; we build+push; pass 2 (below)
 # brings up the full stack. See cdk/README.md "CDK context variables".
-echo "Step 1/5: Pass 1 — creating the ECR repository (CDK)..."
-npx cdk deploy --require-approval never -c imageReady=false "${CDK_CTX[@]}"
-echo "✅ ECR repository created"
+# The check is on stack STATUS, not mere existence: a failed first deploy leaves
+# the stack in ROLLBACK_COMPLETE, which describe-stacks still finds — skipping
+# pass 1 then fails much later with a cryptic "None" repo. Fail early instead.
+STACK_STATUS=$(aws cloudformation describe-stacks --stack-name ClaudeGatewayStack \
+  --region "${DEPLOY_REGION}" --query "Stacks[0].StackStatus" --output text 2>/dev/null || echo "NONE")
+case "$STACK_STATUS" in
+  NONE)
+    echo "Step 1/5: Pass 1 — creating the ECR repository (CDK, first deploy)..."
+    npx cdk deploy --require-approval never -c imageReady=false "${CDK_CTX[@]}"
+    echo "✅ ECR repository created"
+    ;;
+  CREATE_COMPLETE|UPDATE_COMPLETE|UPDATE_ROLLBACK_COMPLETE)
+    echo "Step 1/5: Stack ClaudeGatewayStack already exists (${STACK_STATUS}) — SKIPPING pass 1"
+    echo "          (repo-only) to avoid tearing down the stack. Going straight to build + pass 2."
+    ;;
+  ROLLBACK_COMPLETE)
+    echo "❌ Stack ClaudeGatewayStack is ROLLBACK_COMPLETE (a failed FIRST deploy)."
+    echo "   CloudFormation cannot update a stack in this state. Delete it, then re-run:"
+    echo "     aws cloudformation delete-stack --stack-name ClaudeGatewayStack --region ${DEPLOY_REGION}"
+    echo "     aws cloudformation wait stack-delete-complete --stack-name ClaudeGatewayStack --region ${DEPLOY_REGION}"
+    exit 1
+    ;;
+  *)
+    echo "❌ Stack ClaudeGatewayStack is in state ${STACK_STATUS} — an operation is in"
+    echo "   progress or the stack needs attention. Wait for it to settle (CloudFormation"
+    echo "   console) and re-run."
+    exit 1
+    ;;
+esac
 echo ""
 
 # --- Step 2: Get outputs ---
 echo "Step 2/5: Reading stack outputs..."
 ECR_URI=$(aws cloudformation describe-stacks --stack-name ClaudeGatewayStack \
   --query "Stacks[0].Outputs[?OutputKey=='EcrRepositoryUri'].OutputValue" --output text)
+# Guard: a stack without this output (partial/rolled-back deploy) yields "" or the
+# literal "None" — catching it here beats a cryptic CodeBuild/IAM failure later.
+if [ -z "$ECR_URI" ] || [ "$ECR_URI" = "None" ]; then
+  echo "❌ Could not read EcrRepositoryUri from stack outputs (got: '${ECR_URI:-<empty>}')."
+  echo "   The stack exists but has no ECR output — check its state in the CloudFormation console."
+  exit 1
+fi
 # The repo name the stack actually created, taken from the URI — this is what the
 # CodeBuild IAM policy and docker tags must reference (never assume GATEWAY_NAME
 # alone, so a name mismatch with the stack can't break the push).
@@ -214,7 +247,7 @@ while true; do
   if [ "$STATUS" = "SUCCEEDED" ]; then
     echo "✅ Image built and pushed to ECR"
     break
-  elif [ "$STATUS" = "FAILED" ] || [ "$STATUS" = "STOPPED" ]; then
+  elif [ "$STATUS" = "FAILED" ] || [ "$STATUS" = "STOPPED" ] || [ "$STATUS" = "FAULT" ] || [ "$STATUS" = "TIMED_OUT" ]; then
     echo "❌ Build failed: $STATUS"
     exit 1
   fi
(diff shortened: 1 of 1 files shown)


bluedoors:
Thanks for working on this. This is a bug i'd identfied previously and raise a PR to resolve so this overlaps directly with #266 in the same `deploy.sh` section, so the two changes will need to be reconciled before either merges unchanged. That is a maintainer integration issue rather than a reason to prefer one contribution here.

In terms of this PR, I found two blocking issues.

## The status query fails open

```bash
STACK_STATUS=$(... 2>/dev/null || echo "NONE")
```

This maps every `describe-stacks` failure to the branch that runs the repo-only deployment. A transient throttle, network failure, or refreshed credential can make the status call fail and the following CDK call succeed against a live stack. Since this query is the safety control, it needs to fail closed: treat only an explicit CloudFormation "stack does not exist" response as absence, and stop with the original error for everything else.

The catch-all status branch also gives incorrect recovery advice. `REVIEW_IN_PROGRESS` is a stable change-set-only state left by `--no-execute` or an interrupted deployment; it will not settle by itself. Terminal failures such as `UPDATE_ROLLBACK_FAILED`, `IMPORT_ROLLBACK_FAILED`, `DELETE_FAILED`, and `CREATE_FAILED` also require operator action rather than waiting.

The safest shape is to put the narrow allowlist on the destructive branch: run pass 1 only when the stack is absent or known to contain no deployed resources, and skip pass 1 for every other state. That avoids having to maintain a list of every safe and future CloudFormation status.

## The pass-2 gate rejects routine task-definition replacements

`AWS::ECS::TaskDefinition` is immutable. Any `ContainerDefinitions` change creates a new revision, which CDK reports as `requires replacement` even though this is the normal ECS deployment mechanism.

This local repro needs `2>&1` because the pinned CDK CLI writes its diff to stderr:

```bash
cd claude-apps-gateway/cdk
CTX="-c region=us-east-1 -c gatewayName=claude-gateway -c publicUrl=https://gw.example.com \
 -c zoneName=example.com -c zoneId=Z123456789ABCDEFGHIJK -c ingressCidr=10.0.0.0/8 \
 -c certArn=arn:aws:acm:us-east-1:111122223333:certificate/abc"

npx cdk synth --quiet -c imageReady=true $CTX -c imageTag=latest -o /tmp/p2
NO_COLOR=1 npx cdk diff --template /tmp/p2/ClaudeGatewayStack.template.json \
  -c imageReady=true $CTX -c imageTag=v2 2>&1 |
  grep -Ec '^\[-\]|requires replacement|may be replaced'
```

The results were:

```text
image tag change: 1
no-op:            0
ingressCidr:      0
pass 2 to pass 1: 62
```

I also synthesized the admin API edit from root README section 5. Adding `GATEWAY_ADMIN_WRITE_KEY` and `GATEWAY_ADMIN_READ_KEY` to the container's `secrets` map produces:

```text
[~] AWS::ECS::TaskDefinition Gateway/TaskDef ... replace
 └─ [~] ContainerDefinitions (requires replacement)
```

CPU, memory, environment, and other legitimate task-definition edits have the same result. The current script avoids the image-tag case only because it pins the mutable `:latest` tag, so the task definition itself does not change.

The gate also runs after pass 1, while the destructive failure in #287 happens during pass 1. The status decision is the mechanism that fixes #287; the gate cannot protect its own fail-open path.

I would drop the gate from this PR. If replacement protection is retained, it needs to distinguish routine task-definition revisions from replacement of stateful resources. Matching only `^\[-\]` is not enough because it would allow replacement of RDS, secrets, or ECR. A structured CloudFormation change-set check would be safer than parsing CDK's human-readable output.

## Other changes

The `EcrRepositoryUri` guard and the `FAULT`/`TIMED_OUT` handling are correct and should stay.

Please add a committed regression test for the status decision. The sample already keeps dependency-free shell tests under `claude-apps-gateway/test/`, separate from the Jest suite in `cdk/test/`, and `CLAUDE.md` asks for a test when a deployment trap is fixed. Cases should cover absent, healthy, `ROLLBACK_COMPLETE`, `REVIEW_IN_PROGRESS`, a terminal failure, an in-progress state, and a failed status query.

Finally, please document the new rerun behavior and any retained override in the script's `Usage` block and the CDK README. While you are in the docs, the "its own inline Dockerfile and config" clause at `docs/deployment.md:222` is now stale — `deploy.sh` stamps from `gateway.yaml.template` and uploads the tracked `Dockerfile`. The rest of that note, about the SHA-verified binary, still holds.

bsnehanshu (dismissed):
Withdrawing this review. Deferring to @bluedoors' review above.

scouturier:
Suggested path forward, since this overlaps #266 in the same `deploy.sh` section:

1. Let #266 land first. Its status gate covers the pass-1 fix, and it fails closed on describe-stacks errors (per the review above).
2. Rebase this PR down to the two pieces #266 doesn't have, both already endorsed in review: the `EcrRepositoryUri` empty/`None` guard, and `FAULT`/`TIMED_OUT` as terminal build states.
3. Drop the status `case` (superseded by #266) and the `cdk diff` gate. The gate trips on routine task-definition changes, which are the normal, documented update path.

One housekeeping note: the actual fix for #287 will ship in #266, so the `Fixes #287` here is best changed to a plain reference.

Parking this until #266 merges. The issue you filed was spot-on, and the two guards above are a clean follow-up. Thanks for digging into this!

bluedoors:
Thanks for the careful write-up on this. The `FAULT`/`TIMED_OUT` find is real and is now carried on #266 as 7384fc7fd5a5ba1eb29f4d1ab302e085742279cf, with you as co-author — `buildStatus` has six values and the loop only treated three as terminal, so a faulted or timed-out build polled forever.

Taking #266 for the rest, for three reasons:

- The pass-2 `cdk diff` gate matches `[-] Output NextStep` on a first deploy — pass 1 emits that output and pass 2 drops it — so every fresh install aborts at Step 4/5.
- Any `ContainerDefinitions` edit reports `requires replacement` on `AWS::ECS::TaskDefinition`, which is how ECS revisions work; the gate blocks routine config changes. Only the `:latest` pin hides this today.
- `2>/dev/null || echo "NONE"` maps a throttle, an expired credential, or a network blip to the branch that runs the repo-only deploy against a live stack — the safety control fails open into the bug it is meant to prevent.

#266 resolves the pass-1 destruction with a fail-closed status query, a narrow allowlist on the destructive branch, the same `EcrRepositoryUri` guard, and `test/deploy-helpers.test.sh` covering the status matrix.

Superseded by #266 — I don't have triage rights on this repo to close it here, so a maintainer or you can close it. #287 stays open until #266 merges.

scouturier:
Closing as superseded by #266, which fixes #287 with a fail-closed status query and a committed test for the status decision. Your `FAULT`/`TIMED_OUT` fix is carried there as `7384fc7`, with you as co-author. Thanks for the fix and the detailed write-up, @peepeepopapapeepeepo.

PR #290, opened 2026-08-29
feat(claude-apps-gateway): add update-config.sh and spend-limits.sh ops scripts

### LINKED ISSUE #289: claude-apps-gateway: no tooling for config-only redeploys or spend-limits management (CDK track)
**Sample:** `claude-apps-gateway` (CDK track)

## Gap

Two common day-2 tasks have no tooling in the CDK track, and both are easy to get wrong by hand:

**1. Config-only redeploys.** The gateway config is baked into the container image (README: "Config is baked → the image is per-environment"), so every `gateway.yaml.template` change — model catalog, managed policies, OIDC claim mapping, telemetry — requires an image rebuild and a service roll. The documented path is re-running `deploy.sh`, which drags a full CDK run (and its two passes) into what is logically a config change. There is no supported "rebuild the image and roll the service, touch nothing else" path.

**2. Spend-limits management.** The gateway's [spend-limits Admin API](https://code.claude.com/docs/en/claude-apps-gateway-spend-limits) is documented as raw `curl` calls. In practice the raw API has several traps an operator hits immediately:

- Amounts are whole-number **cents strings**; naive dollar conversion via floats drifts (`12.10` → `1209`).
- `POST` accepts **any string** as `scope.user_id` without validation, but enforcement matches on the OIDC `sub` — a cap keyed on an email silently matches nobody.
- Pagination differs per endpoint: the list and audit endpoints signal with `has_more`, while `/effective` returns **only** `next_page` (no `has_more` field), so a client gating on `has_more` silently truncates to the first page.
- `"0"` is a valid amount that blocks every request for the scope — easy to send by accident.

## Proposal

PR incoming (per CONTRIBUTING, opening this issue first) adding two scripts under `cdk/scripts/`, documented in a new README "Day-2 operations" section:

- `update-config.sh` — stamps the committed template, rebuilds the image through the CodeBuild project `deploy.sh` already created, and rolls the ECS service. Never runs `cdk`; VPC/RDS/ALB/secrets untouched.
- `spend-limits.sh` — a self-contained client for all six Admin API endpoints (`list`/`get`/`set`/`delete`/`effective`/`audit`) plus declarative `apply -f`, with integer-math dollar conversion, an `email:` scope alias that resolves to the real `sub`, correct per-endpoint pagination (`--all`), confirmation prompts on destructive operations, `--dry-run`, and `--json`. Auth via `x-api-key`, bearer, or the session JWT the CLI stores at `claude /login`. bash 3.2 compatible.

Both verified against a live deployment of this stack (ECS Fargate, ap-southeast-7).

peepeepopapapeepeepo:
Closes #289.

Adds two day-2 operations scripts to the CDK track, plus a README "Day-2 operations" section documenting both. No changes to the stack or existing scripts.

## `cdk/scripts/update-config.sh` (116 lines)

Config-only redeploys. The gateway config is baked into the image, so a `gateway.yaml.template` change needs a rebuild + service roll but not a CloudFormation run. The script:

- stamps the committed template via the existing `stamp-config.sh` (same path `deploy.sh` and `setup.sh` use)
- rebuilds and pushes `:latest` through the CodeBuild project `deploy.sh` already created
- forces a new ECS deployment and waits for stability
- never runs `cdk` — VPC, RDS, ALB, and secrets untouched
- refuses to run when the ECS service isn't ACTIVE (mid-teardown), pointing at `deploy.sh` instead

Output style matches `deploy.sh` (`Step N/M`, ✅/❌, closing banner).

## `cdk/scripts/spend-limits.sh` (712 lines)

A self-contained client for the gateway's [spend-limits Admin API](https://code.claude.com/docs/en/claude-apps-gateway-spend-limits), covering all six endpoints — `list`, `get <id>`, `set`, `delete <id>`, `effective`, `audit` — plus declarative `apply -f caps.json` and `resolve <email>`.

Design decisions, each tied to a documented behaviour of the API:

- **Integer-math dollar conversion** (`--usd 12.10` → `"1210"`): the wire format is whole-number cents strings, and float conversion drifts. `--cents`, `--zero`, `--unlimited` map the remaining cases; a final regex refuses to send anything but a cents string or `null`.
- **`email:` scope alias**: the API accepts any string as `scope.user_id` without validation, but enforcement matches on the OIDC `sub` — so `--scope email:alice@…` resolves through `/effective?q=` with an exact, case-insensitive re-filter (the API's `q=` is a substring match) and refuses on ambiguity. A bare `user:` id that looks like an email gets a warning.
- **Per-endpoint pagination**: list and audit page on `has_more`; `/effective` returns only `next_page` (no `has_more` at all). `--all` follows each correctly; without it, a truncated first page prints a hint. `--all` is mutually exclusive with explicit cursors.
- **Credential hygiene**: the auth header is passed to curl via a 0600 temp file (`-H @file`), never argv, so keys/JWTs don't appear in `ps`. Auth resolution: `--key`/`GATEWAY_ADMIN_KEY` (x-api-key) → `--token`/`GATEWAY_BEARER_TOKEN` → the session JWT the CLI stored at `claude /login`.
- **Guard rails**: zero caps and deletes prompt for confirmation (`--yes` to skip); `--dry-run` prints the request without sending; `--json` for machine output; `apply` validates every entry before sending any, so a typo can't half-apply a file.
- **Portability**: bash 3.2 compatible (macOS default shell); `COLUMNS=500` on `column(1)` because util-linux silently drops trailing table columns when non-interactive output exceeds the assumed 80-col width (BSD column doesn't — an easy way to ship a Linux-only display bug from a Mac).

## Testing

- `bash -n` on both scripts; `spend-limits.sh --help` smoke test.
- `cdk` suite: `npm test` — 17/17 (no stack changes).
- `spend-limits.sh` exercised end-to-end against a live deployment of this stack (ECS Fargate): all six endpoints, `--all` pagination, email→sub resolution, sorted `effective`, truncation hints, cursor/`--all` mutual exclusion, and `--dry-run` request shapes.
- `update-config.sh`'s build/roll path is the same CodeBuild + `update-service`/`wait` sequence as `deploy.sh` steps 3/4b, verified live by the same deployment.

By submitting this pull request, I confirm that my contribution is made under the terms of the repository's license.

### DIFF
--- a/claude-apps-gateway/cdk/README.md
+++ b/claude-apps-gateway/cdk/README.md
@@ -243,6 +243,33 @@ it also needs `bootstrapUrl` (→ `<public_url>/user/bootstrap`) in Desktop's ow
 configuration and a `desktop` key on the matching gateway policy — see the root
 [README](../README.md#claude-desktop) and the config reference.
 
+## Day-2 operations
+
+Two helper scripts cover the common post-deploy tasks. Both read `.env` like `deploy.sh` does.
+
+### Update the gateway config without touching infrastructure
+
+```bash
+./scripts/update-config.sh
+```
+
+The config is baked into the container image, so any `gateway.yaml.template` change (models, policies, OIDC claims, telemetry) needs an image rebuild — but not a CloudFormation run. `update-config.sh` stamps the template, rebuilds the image via the existing CodeBuild project, and rolls the ECS service. It never runs `cdk`, so the VPC, RDS, ALB, and secrets are untouched. Use `deploy.sh` instead when the change is to the stack itself (instance sizes, ingress CIDR, VPC endpoints).
+
+### Manage spend limits
+
+```bash
+export GATEWAY_URL=https://claude-gateway.internal.company.com
+./scripts/spend-limits.sh --help
+
+./scripts/spend-limits.sh set --scope organization --usd 50 --period monthly
+./scripts/spend-limits.sh set --scope rbac_group:<group-id> --usd 100 --period monthly
+./scripts/spend-limits.sh set --scope email:alice@company.com --usd 300 --period monthly
+./scripts/spend-limits.sh effective --period monthly --sort spend_desc
+./scripts/spend-limits.sh audit --limit 20
+```
+
+`spend-limits.sh` wraps all six endpoints of the gateway's [spend-limits Admin API](https://code.claude.com/docs/en/claude-apps-gateway-spend-limits) (`list`/`get`/`set`/`delete`/`effective`/`audit`), plus a declarative `apply -f caps.json` and an `email:` scope alias that resolves the address to the OIDC `sub` the API actually keys on. Amounts are converted from dollars with integer math (the wire format is whole-number cents strings), each endpoint's distinct pagination scheme is handled (`--all` follows every page), and destructive operations prompt unless `--yes`. Requires the [`admin:` block](https://code.claude.com/docs/en/claude-apps-gateway-config#admin) in `gateway.yaml` and a network path to the internal ALB; it authenticates with `--key`/`GATEWAY_ADMIN_KEY` (x-api-key), `--token`/`GATEWAY_BEARER_TOKEN`, or falls back to the session JWT the CLI stored at `claude /login`.
+
 ## Configuration reference
 
 All values come from `.env`. The CDK code reads them at deploy time.
--- a/claude-apps-gateway/cdk/scripts/spend-limits.sh
+++ b/claude-apps-gateway/cdk/scripts/spend-limits.sh
@@ -1,0 +1,712 @@
+#!/usr/bin/env bash
+# spend-limits.sh — manage Claude apps gateway per-developer spend caps.
+#
+# Wraps the gateway's Admin API at /v1/organizations/spend_limits, which mirrors
+# Anthropic's public Admin API wire shapes.
+#   https://code.claude.com/docs/en/claude-apps-gateway-spend-limits
+#
+# The gateway serves this API only when gateway.yaml has an `admin:` block, and it
+# sits behind an INTERNAL load balancer — run this from a host with a private
+# network path to the VPC (VPN, Direct Connect, in-VPC).
+#
+# Auth (first match wins):
+#   1. --key / GATEWAY_ADMIN_KEY   → x-api-key      (admin.write_keys = full,
+#                                                    admin.read_keys = GET only)
+#   2. --token / GATEWAY_BEARER_TOKEN → Bearer      (session JWT whose `groups`
+#                                                    claim hits admin.admin_groups)
+#   3. the CLI's own gateway JWT from ~/.claude/.credentials.json
+#
+# Amounts are whole-number strings of USD cents on the wire; this script also
+# accepts dollars via --usd and converts with integer math (no float rounding).
+#
+# Usage: spend-limits.sh <command> [options]     (`--help` for the full list)
+
+set -euo pipefail
+
+# ── Defaults from the environment ────────────────────────────────────────────
+GATEWAY="${GATEWAY_URL:-${GW:-}}"
+ADMIN_KEY="${GATEWAY_ADMIN_KEY:-}"
+BEARER="${GATEWAY_BEARER_TOKEN:-}"
+RAW_JSON=0
+ASSUME_YES=0
+DRY_RUN=0
+FETCH_ALL=0
+
+readonly VALID_PERIODS='daily weekly monthly'
+readonly VALID_SCOPE_TYPES='organization rbac_group user'
+
+TMP_BODY=""
+TMP_HDR=""
+# Must always succeed: as the EXIT trap, a non-zero return here would overwrite the
+# script's real exit status (an unset TMP_BODY made `--help` exit 1).
+cleanup() { [ -n "$TMP_BODY" ] && rm -f "$TMP_BODY"; [ -n "$TMP_HDR" ] && rm -f "$TMP_HDR"; return 0; }
+trap cleanup EXIT INT TERM
+
+die() { printf 'error: %s\n' "$*" >&2; exit 1; }
+note() { printf '%s\n' "$*" >&2; }
+
+need() { command -v "$1" >/dev/null 2>&1 || die "$1 is required but not installed"; }
+
+# ── Help ─────────────────────────────────────────────────────────────────────
+usage() {
+  cat <<'EOF'
+spend-limits.sh — manage Claude apps gateway spend caps
+
+USAGE
+  spend-limits.sh <command> [options]
+
+COMMANDS
+  list                       List configured caps
+  get <spl_id>               Fetch one cap by id
+  set                        Create or replace a cap for {scope, period}
+  delete <spl_id>            Delete one cap
+  effective                  Resolved cap + period-to-date spend per developer
+  audit                      Admin mutation trail (newest first)
+  apply -f <file.json>       Declaratively apply an array of caps
+  resolve <email>            Print the OIDC sub for an email address
+
+GLOBAL OPTIONS
+  -g, --gateway URL   Gateway origin (default: $GATEWAY_URL or $GW)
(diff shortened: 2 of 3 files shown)


bluedoors:
There is useful work here. The integer cents conversion, `jq -n` request construction, endpoint-specific pagination, query-string PII warning, and separation of the auth header from the child `curl` process are all worth keeping.

My recommendation is to split this into two PRs, and to change the shape of the second one:

1. **`update-config.sh`** — keep it as a script in `cdk/scripts/`, with the fixes below. It automates a task this track genuinely lacks.
2. **`spend-limits.sh`** — take the knowledge in it into `workshop/05-spend-caps/` rather than shipping the client. The workshop module already walks the same six endpoints with raw `curl`, and the traps you found are exactly what it is missing; that is where a reader hits them.

To be clear about why, since this is not a judgment on the code quality: `claude-apps-gateway` is a worked example, not a supported artifact, and a 712-line client for an API that lives in a closed-source binary this repository does not own is a maintenance commitment we would be taking on indefinitely — every change to the gateway's admin surface becomes a change here. The findings below are the same argument from the other direction: they are the surface area a client of this kind has to get right, and each one would need a test and an owner. As workshop content the same insight costs a few paragraphs and cannot rot into a wrong answer for someone's production caps.

## What the workshop module should gain

Worth writing up in `workshop/05-spend-caps/README.md`, with a pointer from README section 5:

- Amounts are whole-number strings of USD cents. Converting dollars with floats drifts, so `12.10` can become `1209`.
- `scope.user_id` is the OIDC `sub`. The API accepts any string, but enforcement matches on the `sub`, so a cap keyed on an email address silently matches nobody. Say which behavior you actually observed on `POST` — that is the one part the published docs do not settle.
- Pagination differs per endpoint: the raw list and `/audit` signal with `has_more`, while `/effective` returns only an opaque `next_page`. A client gating on `has_more` truncates `/effective` to the first page.
- `"0"` is a valid amount that blocks every request for the scope, which makes it the most dangerous value to send by accident.
- Resolving a person by email through `/effective` only works after their first request through the gateway: without a `user_ids[]` filter the endpoint lists principals with recorded spend, and `actor.email_address` is null until then. So the one case where you most want to pre-set a cap, a new joiner, is the case it cannot resolve.
- `q=` and `user_ids[]` ride the query string, so they land in ALB and proxy access logs.

## What I found in the client

These are worth fixing in your own copy regardless of where it lives, and they are the concrete reason it is not ready to ship as-is.

### `apply` can partially mutate before rejecting bad input

The pre-validation checks `scope.type`, `period`, and `amount`, but it does not require `scope.user_id` for a user cap or `scope.rbac_group_id` for a group cap, and it does not reject fields that are invalid for the selected scope. I ran the extracted script against this file and all four entries were sent:

```json
[{"scope":{"type":"organization"},"amount":"5000","period":"monthly"},
 {"scope":{"type":"user"},"amount":"100","period":"monthly"},
 {"scope":{"type":"organization","user_id":"bogus"},"amount":"100","period":"daily"},
 {"scope":{"type":"organization"},"amount":"0","period":"daily"}]
```

Entry 0 is applied before the API rejects entry 1, which contradicts the claim that validation prevents half-applied files.

### The zero-cap confirmation does not cover the paths that reach zero

`"0"` blocks every request for a scope, so the confirmation matters, but it is keyed on the `--zero` flag rather than on the amount actually being sent. Two ways past it:

`apply` never prompts at all — the `"amount": "0"` entry above was posted silently.

`--usd` overflows into it. Bash arithmetic wraps, so `--usd 18446744073709551616` produces `"amount": "0"` and no prompt appears even without `--yes`; other large values become unrelated positive amounts (`--usd 92233720368547758` becomes `"9223372036854775800"`). A mistyped or programmatically generated dollar figure can therefore install a total block on the scope without any warning.

### Dry-run

`--dry-run --all` fails with jq exit 5 for `list`, `effective`, and `audit`: pagination gets no response, produces `data: null`, and the renderer cannot iterate it.

`--dry-run resolve` and `--dry-run set --scope email:...` do perform a real `/effective` request, because `resolve_email_to_sub` sets `DRY_RUN=0`. I can see the reasoning in your comment there, and the NOTES section already warns that `email:` and `resolve` reach the same PII-logging endpoint — so this is a help-text gap rather than a behavior defect. The `--dry-run` line promises "send nothing", and it should carve out read-only identity resolution explicitly.

### Auth and transport

The temporary header file keeps the credential out of the child `curl` argv, but credentials passed as `--key` or `--token` remain visible in the parent script's argv, so the guarantee is narrower than stated.

The macOS credentials fallback looks wrong. Anthropic's authentication documentation says macOS credentials live in Keychain; `.credentials.json` is the Linux and Windows path. The `.enterpriseGateway.jwt` shape is also undocumented, so it is not something to depend on without a tested Keychain path.

`api()` treats every status below 400 as success, so a redirect can make `apply` report that it applied caps when no mutation occurred.

`column` is a new dependency for this sample (`jq` already is one via `setup.sh`) and it fails at rendering time if missing, on paths that do not need it — `--json`, `get`, `set`, and `delete`.

## `update-config.sh`

Using `:latest` plus `aws ecs update-service --force-new-deployment` is a valid ECS mechanism. New tasks pull the current image for the existing tag, so this does not have the stale-task behavior described for a plain same-tag CDK redeploy.

It does create an important precondition the script never checks: the active task definition must already reference `:latest`. The stack's default is the pinned claude version (`bin/app.ts`: `imageTag ?? claudeVersion`, currently `2.1.229`), so a deployment made through the documented manual CDK path references `:2.1.229`. Against that stack this script pushes an unused `:latest` image and then forces the service to redeploy the same old versioned image — a config update that reports success and changes nothing. Please inspect the active task definition and fail clearly unless it uses the mutable tag.

The mutable tag also gives up config-addressable rollback and is incompatible with ECR tag immutability. The script and README should state that tradeoff and explain how to roll back. A hashed-tag workflow would need to update the task definition through CDK or another explicitly documented mechanism.

The existing `ACTIVE` check does not verify the other prerequisites. An Option B deployment can have an active service but no `claude-gateway-build` project or build bucket, so the script reaches a raw S3 or CodeBuild error. Check both resources before stamping and uploading.

The region fallback should copy `deploy.sh`'s `|| true` and `:?` validation. As written, a missing AWS profile region exits during assignment under `set -e`, and missing `.env` values surface as `unbound variable` before `stamp-config.sh` can print its clearer messages.

Step 2 duplicates the buildspec, four uploads, and CodeBuild polling loop from `deploy.sh`, including a second copy of the `FAULT`/`TIMED_OUT` fix. Please extract the shared build path or otherwise give it one tested implementation.

The script also stages whichever local `claude` binary it finds, so a command described as a config-only update can change the gateway binary at the same time. Either verify the expected pinned binary or document that the whole image payload may change.

There is also an integration conflict with #266: that PR changes the build bucket to `claude-gateway-build-${ACCOUNT_ID}-${DEPLOY_REGION}`. With the current hard-coded unsuffixed name, this script either fails when the old bucket is absent or uploads to stale source when it remains. The bucket should come from the CodeBuild project's configured source rather than being reconstructed independently.

scouturier:
Thanks for the contribution. After review with the team, we don't expect to merge this PR. Summary:

**`spend-limits.sh`**: we keep out of the repo any content that could drift. The spend-limits Admin API is a product surface Anthropic owns, so hosting a client for it here (or documenting the API's behavior in the workshop or a README) creates a second source of truth that goes stale as the product evolves. We point readers at the published spend-limits docs instead. This is not a reflection of the code itself.

**`update-config.sh`**: superseded by #266. `deploy.sh` already covers config-only redeploys end to end (config stamp, CodeBuild rebuild, then an ECS roll via `--force-new-deployment`). The only thing that made re-running it unsafe is the pass-1 bug #266 fixes, so a second copy of that build path isn't worth its maintenance cost.

We'll close this PR since #266 resolves the remaining. Happy to answer questions.

scouturier:
Closing per the review above. `spend-limits.sh` is out of scope for the repo regardless.

`update-config.sh`: the config-only redeploy path is landing in `deploy.sh` via #266 (re-run safety tracked in #287), so we won't take a second script for it. Thanks again for the work here!

PR #291, opened 2026-08-31
chore(claude-apps-gateway): bump pinned Claude Code to 2.1.274; pin stopTimeout above the SIGTERM drain

bluedoors:
## What

Bumps the `claude-apps-gateway` worked example's pinned Claude Code version from **2.1.229 to 2.1.274** (`CLAUDE_VERSION` in `cdk/scripts/setup.sh`, `claudeVersion` in `cdk/bin/app.ts`), adds **`bedrock:CountTokens`** to the task role, **pins the container `stopTimeout` above the gateway's SIGTERM drain window**, and reconciles the documentation the intervening releases invalidated.

Opened as a 2.1.251 bump and retargeted as releases landed. The current target is driven by **2.1.274**, the most server-relevant release for this example since 2.1.271.

## Why: the server-side gates

| Release | Change | Why this example cares |
|---|---|---|
| **2.1.232** | `desktop` block went from 11 hand-listed keys to Claude Desktop's full settings schema; empty `match.groups` / `admin.admin_groups` and malformed `email_domain` now fail at boot | This example ships a `desktop` block, and the docs asserted the 11-key set as fact in four places |
| **2.1.233** | `400`/`413` from a cloud upstream carries the upstream's own message | Legible Bedrock failures |
| **2.1.260** | Aborted requests counted via Bedrock's free `CountTokens` API | **New IAM action** — see below |
| **2.1.261** | Client IP fixed when a proxy appends a port to `X-Forwarded-For`; `403` on an unreadable access-list entry | This example sits behind an ALB |
| **2.1.265** | OTLP relay no longer pauses forwarding for 30s after a rejected payload; direct-to-collector export | This example ships the relay and an ADOT sidecar |
| **2.1.268** | `pricing:` rates reach signed-in clients; empty-`allow_cidrs` startup warning | `pricing:` is no longer spend-meter-only; the warning fires on the shipped config |
| **2.1.271** | `pricing.multiplier` ceiling `1` -> `10` | Makes the residency premium a one-liner (#308) |
| **2.1.273** | `allowManagedMcpServersOnly`, `deniedMcpServers`, `disableClaudeAiConnectors` set via MDM no longer ignored when server-managed settings are present | **Always** true in this topology, so those keys were silently inert on exactly the machines this example targets. The "pair it with the allowManaged*Only locks" advice in `gateway.yaml.template` now says so |
| **2.1.274** | **SIGTERM no longer cuts every open stream** — in-flight requests get up to 25s (`CLAUDE_GATEWAY_DRAIN_TIMEOUT_MS`) | The headline: below this, every ECS rolling deploy severs live streaming responses, the failure the 3600s ALB idle timeout exists to prevent |
| **2.1.274** | `store.connect_timeout_seconds` (default 5s), boot error naming `store.postgres_url`, first Postgres connection retried 3x | RDS is in private subnets here and can answer a few seconds late; that was a boot crash-loop |
| **2.1.274** | Spend-limit checks take 1 DB round trip instead of 4; unhandled rejection when Postgres drops a connection mid-spend-check fixed | Both on the spend-meter path README section 5 documents |
| **2.1.274** | Warning + startup line for the 256-concurrent-upstream-request limit | Capacity signal against `DESIRED_COUNT` (2). The boot line names the knob the changelog omits: `BUN_CONFIG_MAX_HTTP_REQUESTS` |

## New: `stopTimeout` pinned above the drain window

The 2.1.274 drain fix is inert unless ECS lets the container live long enough to use it. ECS SIGKILLs at `stopTimeout`, and **nothing in this repo set it** — leaving the ECS default of 30s, which covers a 25s drain but only just, and silently breaks for anyone who lowers it. Both tracks now pin **40s**:

- CDK: a property override, because `ApplicationLoadBalancedFargateService`'s `taskImageOptions` exposes no `stopTimeout`.
- `setup.sh`: in the task-def JSON.
- A CDK test asserts the container **named `web`** carries a value above 25, so the index-based override cannot drift onto the ADOT sidecar if container order changes.
- `CLAUDE.md` records the constraint: raise `CLAUDE_GATEWAY_DRAIN_TIMEOUT_MS` and `stopTimeout` together, never one.

## Pinning `latest` — a deliberate exception, stated in the docs

2.1.274 was `latest` with nothing above it, which breaks this repo's own "never pin the newest release" rule (large releases have repeatedly needed a same-week follow-up: 2.1.266<-265, 2.1.270<-269, 2.1.272<-271). Taken anyway, because the defect it fixes is in **this example's own deployment shape** and no earlier release fixes it. `docs/upstream-watch.md` states the exception and its reasoning rather than contradicting itself, and tells anyone pinning fresh to prefer 2.1.275+ if one exists.

## The `bedrock:CountTokens` action, and why it is documented as a nicety

Added to the task role in the CDK stack, `setup.sh`, and the asserted policy in the CDK tests. Two findings kept it from being written up as a correctness gate:

- **The gateway falls back.** It logs `bedrock CountTokens failed, using a max_tokens:1 request for the aborted-request token count` once per upstream and meters through that probe instead. Nothing goes uncounted without the grant; you pay for a minimal request.
- **Bedrock mostly does not support it yet.** `CountTokens` takes only a bare foundation-model id (the gateway strips the `global.`/`us.` prefix itself), and of this catalog only `anthropic.claude-haiku-4-5-20251001-v1:0` supports it; `anthropic.claude-opus-5` and `anthropic.claude-sonnet-5` return `ValidationException: The provided model doesn't support counting tokens`. Probed directly against Bedrock, us-east-1.

## Also in here

- **README** gains a developer-floor table and an explicit **"one client version to skip: 2.1.265"** — that release made the undocumented `CLAUDE_CODE_USE_GATEWAY` variable force gateway sign-in on its own, breaking every request on machines that set it alongside an API key, `apiKeyHelper`, or custom auth headers. Restored in 2.1.266.
- **`docs/upstream-watch.md`** gains the 2.1.260/261/265/268/271/273/274 tripwires and observations from the live runs, including one that changes how you read the audit trail: **`client_ip` is the load balancer, not the developer** (`session.mint` logged `10.20.10.186`, an ALB ENI, while the signing-in laptop was on the VPN) because `listen.trusted_proxies` is unset. Anything keyed on client IP sees one address for the whole fleet until it is set.

## Validation

Validated **live end-to-end on 2.1.274** against the ECS Fargate track, as an in-place image swap on the running service. The baked config was byte-identical to the previous run, so the binary and the new `stopTimeout` were the only variables.

| Check | Result |
|---|---|
| Release artifact provenance | GPG signature on `manifest.json` good against the pinned key; `linux-x64` (`15e2d051...`) and `darwin-arm64` (`3509913f...`) SHA-256 match the signed manifest |
| Version gate, probed not assumed | `CLAUDE_GATEWAY_DRAIN_TIMEOUT_MS` and `store.connect_timeout_seconds` return **zero `strings` hits on 2.1.272 and 2.1.273** and are present in 2.1.274, so the gate is exactly 2.1.274 |
| `stopTimeout` in the synthesized template | `web` carries `StopTimeout: 40`; the ADOT sidecar is untouched |
| Service rollout | 2/2 tasks, both ALB target groups healthy, the out-of-band `admin-ui` binding preserved |
| Gateway boot, both replicas | `config.load` clean, Bedrock upstream, `telemetry relay: 1 destination(s), signals enabled: metrics,logs`, `managed settings: configured` |
| New 2.1.274 startup line | `upstream requests: at most 256 at once per process (set BUN_CONFIG_MAX_HTTP_REQUESTS to change)` |
| In-VPC probes | DNS resolves to private addresses only; `/healthz` 200; `/readyz` 200 |
| Client-side probes over the VPN | TLS chain verifies (`Verify return code: 0`); `/healthz`, `/readyz` and `/admin/` all 200 |
| Desktop opt-in | `/user/bootstrap` unauthenticated returns **401**, not 404 |
| Sign-in | Browser SSO completed; gateway logged `session.mint` `result:success` |
| Inference | One `/v1/messages` call per model in `availableModels` — `claude-opus-5`, `claude-sonnet-5`, `claude-haiku-4-5` — all **200**, driven with a bearer token to sidestep the CLI's silent `--model` fallback |
| Spend meter, after 2.1.274 rewrote the spend-check query path | Exact: a controlled Opus 5 call predicted a daily total of `30.486060` cents at the configured `multiplier: 1.1`; Postgres read **`30.48606`**. The list-price figure would have been `30.447510` |
| Local suites | 18/18 CDK Jest (including the new `stopTimeout` assertion), 9/9 `stamp-config`, 19/19 `setup-helpers`, `cdk synth` OK, `shellcheck` clean |

**Not verified:** the SIGTERM drain itself. Confirming it needs an immediate SIGTERM to a task serving an open stream, and a rolling deploy does not deliver one promptly enough to catch a stream that finishes in ~30s. The `stopTimeout` change is asserted in the CDK test and present in the deployed task definition; the drain behaviour is taken from the changelog and from the version gate probed above.

## Follow-up: version prose consolidated (`1301e69`)

The README had accumulated **25 version references, ~20 of them archaeology** — `requires ≥ 2.1.198` for cross-upstream failover, `≥ 2.1.203` for `/user/bootstrap`, `≥ 2.1.227` and `≥ 2.1.232` scattered through the `desktop` key table, and the seven-row "why this pin" table this PR originally added. The example pins one version, so the pin satisfies every one of them; a reader deploying it learns nothing, and a bump has to edit the same fact in several files.

**README §4 now leads with one statement** — *"This example is built and validated on Claude Code `2.1.272`"* — says outright that there are no per-feature gates below it, and points at `docs/upstream-watch.md` for anyone adapting the example onto an older gateway they already run.

What survives is the axis a single pin **can't** absorb: the client floor. The pin governs the container, not the fleet's laptops, so the developer-facing versions stay (now a table): 2.1.195 minimum, the gateway-session fixes in 2.1.237/247/248/251/260, the `≥ 2.1.270` needed to honour a `multiplier` above 1, and the one release to skip (2.1.265).

`docs/upstream-watch.md` — already the doc whose job is version tracking — gains a **"Why this pin"** section with the moved table and the 2.1.272-not-2.1.271 reasoning. Its tripwire list already carried every per-feature gate removed from the README except 2.1.229's SSE keepalive pings, now added. Checked that no gate was lost in the move and that no cross-doc link pointed at the renamed §4 heading.

Version references after the change: README 13 (all client-axis or the pin itself), `upstream-watch.md` 70. A future bump touches the pin line and that one file.

## Earlier commit: the client half of the multiplier gate (`262b777`)

The pin this PR lands carries 2.1.271's `pricing.multiplier` ceiling of `10`, but *honouring* a markup is a client capability gated one release lower:

| Side | Floor |
|---|---|
| Gateway server (this pin) | **≥ 2.1.271** to accept a `multiplier` above `1` at all |
| Developer's CLI | **≥ 2.1.270** to honour it |

Below 2.1.270 the developer's `/cost` reports list price while the gateway's caps and spend records already carry the markup. Observed at boot on the live 2.1.272 deployment: *"Claude Code clients older than v2.1.270 ignore a multiplier above 1 and show costs without the markup."* A discount below `1` has no client floor.

Only the version-gate half is here — the README's version notes and `upstream-watch.md`'s client-side-gates section, which this PR already owns. The `pricing:` block itself is documented in #308.

### DIFF
--- a/claude-apps-gateway/README.md
+++ b/claude-apps-gateway/README.md
@@ -109,7 +109,13 @@ Both the gateway server (Linux binary) and each developer's Claude Code CLI must
 
 Developers can update with `claude update`. The gateway server uses the same binary, downloaded from the Claude Code release page and packaged into a container image.
 
-Some gateway behaviour is version-gated: v2.1.198 added cross-upstream failover on `404` and the `anthropicAws` (Claude Platform on AWS) provider — earlier gateway builds reject that provider at boot; v2.1.203 added the Claude Desktop bootstrap endpoint (`/user/bootstrap`); v2.1.227 added the `desktop` block's `chatTabEnabled` and `chatAdvancedFileAnalysisEnabled` keys, the `oidc.use_proxy` flag, and the `pricing:` block. The worked example in this repo pins **2.1.229**, which adds SSE keepalive pings on streaming responses so long thinking pauses don't trip an idle timeout on the Bedrock upstream — this example still raises the ALB idle timeout to 3600s as well, since the ALB has to outlast the stream either way. The pin also carries the earlier fix that prices Bedrock application-inference-profile ARNs and other config-mapped upstream model IDs at the configured model's rates, directly relevant to this example's `global.anthropic.*` inference profiles. See [`docs/upstream-watch.md`](docs/upstream-watch.md) for a checklist to stay across gateway releases.
+Some gateway behaviour is version-gated: v2.1.198 added cross-upstream failover on `404` and the `anthropicAws` (Claude Platform on AWS) provider — earlier gateway builds reject that provider at boot; v2.1.203 added the Claude Desktop bootstrap endpoint (`/user/bootstrap`); v2.1.227 added the `desktop` block's `chatTabEnabled` and `chatAdvancedFileAnalysisEnabled` keys, the `oidc.use_proxy` flag, and the `pricing:` block; v2.1.229 added SSE keepalive pings on streaming responses so long thinking pauses don't trip an idle timeout on the Bedrock upstream (this example still raises the ALB idle timeout to 3600s as well, since the ALB has to outlast the stream either way), and carries the earlier fix that prices Bedrock application-inference-profile ARNs and other config-mapped upstream model IDs at the configured model's rates, directly relevant to this example's `global.anthropic.*` inference profiles.
+
+The worked example in this repo pins **2.1.251**. Two server-side gates drove that pin past 2.1.229: **v2.1.232** widened the `desktop` block from 11 hand-listed keys to Claude Desktop's full settings schema (and added `disabledBuiltinTools`, `coworkEgressAllowedHosts`, `managedMcpServers`, plus boot-time rejection of empty `match.groups` / `admin.admin_groups` entries and malformed `email_domain` values — previously those silently matched no one or granted admin access), and **v2.1.233** made `400`/`413` responses from a cloud upstream carry the upstream's own message.
+
+**Developers benefit from being newer than the floor, independently of the pin.** The gateway-facing client fixes worth telling your fleet about: v2.1.237 and v2.1.248 fixed prompt caching on gateway sessions (the latter a roughly hourly cache miss caused by an OAuth token refresh); v2.1.248 fixed `/login` to a gateway hanging when the managed-settings approval dialog was required, and v2.1.251 stopped that dialog re-appearing on every re-sign-in and reduced it to only the settings that changed; v2.1.247 fixed first-run setup exiting with "Unable to connect to Anthropic services" when managed settings force gateway sign-in and Anthropic's own endpoints are unreachable — the normal case on a restricted-egress network. v2.1.251 also adds a **Spend limit** bar to `/usage` for developers behind a gateway with spend limits configured; that one needs v2.1.251 on the developer's machine but nothing newer than v2.1.225 on the server.
+
+See [`docs/upstream-watch.md`](docs/upstream-watch.md) for a checklist to stay across gateway releases.
 
 ### 5. Device management (for pushing settings to developers)
 
@@ -202,7 +208,7 @@ that explicitly opt in. `/user/bootstrap` — the endpoint Desktop fetches its c
 returns `404` unless the matching policy carries a `desktop` key. An empty `desktop: {}`
 is enough to opt a policy in, and a `desktop` key on the `match: {}` base layer opts in
 every policy that inherits it. Requires the gateway server on **v2.1.203+** (this example
-pins 2.1.229). Pair it with `bootstrapUrl` on the client side — see
+pins 2.1.251). Pair it with `bootstrapUrl` on the client side — see
 ["How developers connect"](#claude-desktop).
 
 ```yaml
@@ -225,9 +231,13 @@ which fans out to your destinations. Keys with no Desktop equivalent are omitted
 `hooks` and scoped rules like `Bash(npm *)`; if a restriction matters for Desktop, express
 it as a bare tool name.
 
-The `desktop:` block itself holds only the Desktop-specific gates that have no CLI
+The `desktop:` block itself holds the Desktop-specific settings that have no CLI
 equivalent. Every key is optional (Desktop applies its own default for anything omitted),
-and unknown keys **fail gateway boot**:
+and unknown keys **fail gateway boot**. Since gateway **2.1.232** the block accepts every
+released Claude Desktop setting and is validated against Desktop's own schema, so the table
+below is the useful subset for a gateway deployment, not the whole accepted set — write any
+key from Claude Desktop's [managed configuration reference](https://claude.com/docs/third-party/claude-desktop/configuration)
+as a flat key name:
 
 | Key | Effect |
 |---|---|
@@ -239,10 +249,12 @@ and unknown keys **fail gateway boot**:
 | `isLocalDevMcpEnabled` | Allow locally defined MCP servers |
 | `disableAutoUpdates`, `autoUpdaterEnforcementHours` | Auto-update policy |
 | `banner` | Persistent banner in the app: `enabled`, `text`, `backgroundColor`, `textColor`, `linkUrl` |
+| `disabledBuiltinTools` | Extra tools to disable — **unioned** with the list derived from bare-name `permissions.deny`, so it can only disable more, never re-enable. Needs gateway **≥ 2.1.232** |
+| `coworkEgressAllowedHosts` | Cowork egress allowlist — **replaces** the list derived from `sandbox.network.allowedDomains`. Needs gateway **≥ 2.1.232** |
+| `managedMcpServers` | The only way to push MCP servers from a policy: the gateway rejects `mcpServers` inside a `cli` block at boot, but Desktop clients can receive them here. Needs gateway **≥ 2.1.232** |
 
-The table above is the full accepted key set as of 2.1.229. Three traps bite here — a
-banner that renders nothing, a silently missing Chat tab, and an unknown key that
-crash-loops the ECS task; all three are written up in
+Three traps bite here — a banner that renders nothing, a silently missing Chat tab, and an
+unknown key that crash-loops the ECS task; all three are written up in
 [`docs/gotchas.md`](docs/gotchas.md#20-three-ways-the-desktop-block-itself-bites).
 
 If you don't deploy Claude Desktop, leave `desktop` out entirely — `/user/bootstrap` then
--- a/claude-apps-gateway/cdk/gateway.yaml.example
+++ b/claude-apps-gateway/cdk/gateway.yaml.example
@@ -193,7 +193,11 @@ managed:
       #
       # `desktop: {}` on its own is a complete opt-in. The keys below are optional
       # feature gates with no CLI equivalent; each one omitted keeps Desktop's own
-      # default, and an UNKNOWN key here fails gateway boot. The rest of the response
+      # default. As of gateway 2.1.232 this block accepts every released Claude
+      # Desktop setting and is validated at boot against Desktop's OWN schema, so
+      # an UNKNOWN key — or a recognized key with a value Desktop would reject, a
+      # legacy alias, or a key the gateway computes itself (inference connection,
+      # model list, OTLP relay) — fails gateway boot. The rest of the response
       # comes from the cli: block above — availableModels, bare tool-name
       # permissions.deny entries, sandbox.network.allowedDomains, and an OTLP endpoint
       # when telemetry.forward_to is set. hooks and scoped rules like Bash(npm *) have
@@ -210,6 +214,14 @@ managed:
       #   # these keys need gateway >= 2.1.227 — on an older pin they fail boot.
       #   chatTabEnabled: true
       #   chatAdvancedFileAnalysisEnabled: false  # no effect if permissions.deny blocks Bash
+      #   # These three need gateway >= 2.1.232. disabledBuiltinTools is UNIONED with
+      #   # the list derived from bare-tool-name permissions.deny above (so it can only
+      #   # disable more, never re-enable); coworkEgressAllowedHosts REPLACES the list
+      #   # derived from sandbox.network.allowedDomains. managedMcpServers is the only
+      #   # way to push MCP servers from a policy — the gateway rejects mcpServers
+      #   # inside the cli: block above, at boot.
+      #   disabledBuiltinTools: [WebSearch]
+      #   coworkEgressAllowedHosts: ["*.internal.example.com"]
       #   isDesktopExtensionEnabled: true
       #   isDesktopExtensionSignatureRequired: true
       #   isLocalDevMcpEnabled: false
--- a/claude-apps-gateway/cdk/gateway.yaml.template
+++ b/claude-apps-gateway/cdk/gateway.yaml.template
@@ -200,7 +200,11 @@ managed:
       #
       # `desktop: {}` on its own is a complete opt-in. The keys below are optional
       # feature gates with no CLI equivalent; each one omitted keeps Desktop's own
-      # default, and an UNKNOWN key here fails gateway boot. The rest of the response
+      # default. As of gateway 2.1.232 this block accepts every released Claude
+      # Desktop setting and is validated at boot against Desktop's OWN schema, so
+      # an UNKNOWN key — or a recognized key with a value Desktop would reject, a
+      # legacy alias, or a key the gateway computes itself (inference connection,
+      # model list, OTLP relay) — fails gateway boot. The rest of the response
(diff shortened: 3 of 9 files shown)


claude-apps-gateway/README.md
```diff
@@ -225,24 +255,30 @@ which fans out to your destinations. Keys with no Desktop equivalent are omitted
...
+| `chatTabEnabled` | Show or hide the Chat tab — **hidden unless set `true`** |
+| `chatAdvancedFileAnalysisEnabled` | Let Claude run code in a local sandbox from the Chat tab to analyse attached files it can't read natively (spreadsheets, presentations). Off unless set `true`; no effect when the policy's `permissions.deny` disables `Bash` |
 | `isDesktopExtensionEnabled`, `isDesktopExtensionSignatureRequired` | Desktop extension loading and signature checks |
 | `isLocalDevMcpEnabled` | Allow locally defined MCP servers |
 | `disableAutoUpdates`, `autoUpdaterEnforcementHours` | Auto-update policy |
 | `banner` | Persistent banner in the app: `enabled`, `text`, `backgroundColor`, `textColor`, `linkUrl` |
+| `disabledBuiltinTools` | Extra tools to disable — **unioned** with the list derived from bare-name `permissions.deny`, so it can only disable more, never re-enable |
+| `coworkEgressAllowedHosts` | Cowork egress allowlist — **replaces** the list derived from `sandbox.network.allowedDomains` |
+| `managedMcpServers` | The only way to push MCP servers from a policy: the gateway rejects `mcpServers` inside a `cli` block at boot, but Desktop clients can receive them here |
```
scouturier: Backwards. Per the [config reference](https://code.claude.com/docs/en/claude-apps-gateway-config#mcp-servers-in-a-policy), MCP servers reach Claude Code clients via `managedMcpServers` in the policy's **`cli`** block (2.1.259+, so fine on this pin). Only the `.mcp.json` spelling `mcpServers` is rejected there. This `desktop` key is Desktop-only and takes an **array**, not the CLI's object.

claude-apps-gateway/cdk/gateway.yaml.template
```diff
@@ -217,6 +225,14 @@ managed:
       #   # these keys need gateway >= 2.1.227 — on an older pin they fail boot.
       #   chatTabEnabled: true
       #   chatAdvancedFileAnalysisEnabled: false  # no effect if permissions.deny blocks Bash
+      #   # These three need gateway >= 2.1.232. disabledBuiltinTools is UNIONED with
+      #   # the list derived from bare-tool-name permissions.deny above (so it can only
+      #   # disable more, never re-enable); coworkEgressAllowedHosts REPLACES the list
+      #   # derived from sandbox.network.allowedDomains. managedMcpServers is the only
```
scouturier: Same correction as `README.md:278`: `managedMcpServers` in a `cli` block is the supported path for CLI clients (2.1.259+). The gateway rejects only `mcpServers`. As written, a reader who wants MCP servers on their CLI fleet puts them here and the clients never receive them.

scouturier (commented):
Verified the load-bearing parts: 18/18 CDK tests pass, the synthesized template puts `StopTimeout: 40` on `web` with the sidecar untouched, and every release claim from 2.1.232 to 2.1.274 matches the changelog verbatim.

Inline notes below. One is a flat factual error (`managedMcpServers`, in three files); the rest are `upstream-watch.md` drifting into restating product behaviour the published docs already own.

**Target end-state:** land the bump, the `CountTokens` grant and the `stopTimeout` pin as they are; fix the `managedMcpServers` direction in all three files; note the distance from `stable`; and let `upstream-watch.md` defer to the published docs rather than shadow them.

claude-apps-gateway/cdk/gateway.yaml.example
```diff
@@ -210,6 +218,14 @@ managed:
       #   # these keys need gateway >= 2.1.227 — on an older pin they fail boot.
       #   chatTabEnabled: true
       #   chatAdvancedFileAnalysisEnabled: false  # no effect if permissions.deny blocks Bash
+      #   # These three need gateway >= 2.1.232. disabledBuiltinTools is UNIONED with
+      #   # the list derived from bare-tool-name permissions.deny above (so it can only
+      #   # disable more, never re-enable); coworkEgressAllowedHosts REPLACES the list
+      #   # derived from sandbox.network.allowedDomains. managedMcpServers is the only
```
scouturier: Same as the template at line 231. Keep the two files in step when you fix it.

claude-apps-gateway/docs/upstream-watch.md
```diff
@@ -10,13 +10,78 @@ Run it **before each version bump** and **when a gateway release ships**.
...
+| **2.1.274** | A warning when a replica exceeds the 256 requests it sends upstream at once, plus a startup log line showing the limit | Capacity signal worth knowing against `DESIRED_COUNT` (2 here). The boot line names the knob the changelog doesn't: `upstream requests: at most 256 at once per process (set BUN_CONFIG_MAX_HTTP_REQUESTS to change)` |
+
+**Why the pin is 2.1.274, which was `latest` at the time — a deliberate exception.** The rule
+this section otherwise follows is *never pin the newest release*: a large release has repeatedly
+needed a same-week follow-up (2.1.266←265, 2.1.270←269, 2.1.272←271), so pinning head leaves no
+successor to fix it. 2.1.274 is a large release and had nothing above it when this pin was
+chosen, so it breaks that rule on purpose: it repairs a defect in **this example's own
+deployment shape** — a rolling ECS deploy cutting live streams — and no earlier release does.
+The trade is stated rather than hidden; if you are pinning fresh and a 2.1.275+ exists, prefer
```
scouturier: This is currently a no-op: no 2.1.275+ exists yet (probed 275/276/277, all 404), so the head-pin exception is live. Worth naming the distance from the promoted channel too: `downloads.claude.ai/claude-code-releases/stable` returns **2.1.267**, seven releases back. A reader copying this sample inherits the exception and cannot otherwise tell.

claude-apps-gateway/docs/upstream-watch.md
```diff
@@ -10,13 +10,78 @@ Run it **before each version bump** and **when a gateway release ships**.
...
+| **2.1.274** | `store.connect_timeout_seconds` (default 5s), a boot error naming `store.postgres_url`, and the first Postgres connection retried 3× before exit | RDS sits in private subnets here and can be reachable a few seconds after the task starts; that used to be a boot crash-loop |
+| **2.1.274** | Spend-limit checks take one database round trip instead of four, and an unhandled promise rejection when Postgres drops a connection mid-spend-check is fixed | Both are on the spend-meter path this example documents in README §5 |
+| **2.1.274** | A warning when a replica exceeds the 256 requests it sends upstream at once, plus a startup log line showing the limit | Capacity signal worth knowing against `DESIRED_COUNT` (2 here). The boot line names the knob the changelog doesn't: `upstream requests: at most 256 at once per process (set BUN_CONFIG_MAX_HTTP_REQUESTS to change)` |
```
scouturier: `BUN_CONFIG_MAX_HTTP_REQUESTS` is a runtime internal, not a supported tuning surface, and the note says the changelog omits it. Suggest keeping the 256 capacity signal and dropping the knob.

claude-apps-gateway/docs/upstream-watch.md
```diff
@@ -53,15 +123,114 @@ exceeds our pin, we're behind on that feature. The ones we already track:
...
+- `forward_user_identity` on an `anthropic` upstream (per-user attribution at a proxy
+  *behind* the gateway; not applicable to a Bedrock upstream) — **requires ≥ 2.1.233**
+- **Graceful shutdown: in-flight requests get up to 25s on SIGTERM, tunable with
+  `CLAUDE_GATEWAY_DRAIN_TIMEOUT_MS` — added in 2.1.274.** Below this the gateway cut every open
+  stream the moment ECS sent SIGTERM, so a rolling deploy killed live streaming responses. Two
+  things to keep in step with it, both now pinned in the CDK stack and `setup.sh`: the container
+  **`stopTimeout` must outlast the drain window** (ECS SIGKILLs at `stopTimeout`; its default of
+  30s only just covers 25s, so this example sets **40s** explicitly and a CDK test asserts it is
+  above 25), and if you raise `CLAUDE_GATEWAY_DRAIN_TIMEOUT_MS` you must raise `stopTimeout` with
+  it (Fargate's ceiling is 120s). `strings` on the 2.1.272 and 2.1.273 binaries returns **0 hits**
```
scouturier: Good diligence for your own confidence, but publishing `strings` as the way to establish a version gate teaches readers to reverse-engineer the binary rather than read the changelog. The 2.1.274 gate stands on the changelog entry alone.

claude-apps-gateway/docs/upstream-watch.md
```diff
@@ -91,16 +260,30 @@ live on 2.1.229: with the `desktop` block enabled, Claude Desktop showed the Cha
   `cli` key newer than the pin fails **at gateway boot** or is rejected server-side
   at `/v1/messages`. Confirm every model in `availableModels` and every `cli` key is
   known to the pinned version. (See the README's Claude Code version prerequisite.)
+  The `cli` schema is closed apart from `env`, `pluginConfigs`, and keys nested under
+  `permissions`, which pass through for newer clients. Settings this pin newly *can*
+  deliver, none of them shipped in our block yet: `modelPricing`, `modelPicker`,
+  `promptCacheTtl`, `subagentPromptCacheTtl` (2.1.243), `feedbackDrafts` and the extended
+  `spinnerTipsOverride` entry shape (2.1.247). Probe each before shipping it — none is
```
scouturier: These keys are undocumented by your own admission, which makes them the moving target this repo tries not to publish. A sample is where readers look for the supported path; suggest cutting the inventory.

claude-apps-gateway/docs/upstream-watch.md
```diff
@@ -53,15 +123,114 @@ exceeds our pin, we're behind on that feature. The ones we already track:
...
 strictly** — an unknown key fails gateway boot (crash-loops the container on ECS), so a
-key copied from the docs page for a newer release takes the deployment down. As of
-`2.1.229` the accepted keys are exactly: `modelDiscoveryEnabled`, `coworkTabEnabled`,
-`isClaudeCodeForDesktopEnabled`, `chatTabEnabled`, `chatAdvancedFileAnalysisEnabled`,
-`isDesktopExtensionEnabled`, `isDesktopExtensionSignatureRequired`, `isLocalDevMcpEnabled`,
-`disableAutoUpdates`, `autoUpdaterEnforcementHours`, `banner`. Verify a new key against the
-pinned binary before shipping it:
+key copied from the docs page for a newer release takes the deployment down.
+
+**2.1.232 changed the shape of this check.** Through `2.1.231` the gateway accepted a fixed
```
scouturier: This bullet list restates the config reference's own boot-failure list nearly verbatim (I diffed them), so it goes stale the moment that page changes. Link it as the authority and keep a one-line summary.

PR #294, opened 2026-09-01
Fix Fable 5.1 getting-started: data-retention modes and remove commercial Mantle

DanyStinson:
## Summary

Corrects two factual issues in the Claude Fable 5.1 getting-started notebook, both verified live against Amazon Bedrock at launch (2026-09-01).

### 1. Data retention modes
The notebook stated access requires `aws_review`. Verified on an enforced account: the Covered Model gate accepts **both `aws_review` and `provider_data_share`**, and rejects `none` and `default`. A customer already on `provider_data_share` (Fable 5's mode) does not need to switch. Updated the What-you'll-learn list, Access prerequisites, section 2, the migration checklist, and the summary accordingly.

### 2. Bedrock Mantle is not a commercial path
The notebook presented the Anthropic Messages API on the Bedrock Mantle endpoint as a supported way to call the model. Mantle is **not available for Fable 5.1 in commercial Regions** (runtime-only at commercial launch); on Bedrock it is offered only in AWS GovCloud (US). The Mantle endpoint returns a deterministic 404 for Fable 5.1 in commercial Regions. Removed the Mantle model ID, the Mantle invocation cell, and the stale 'us-east-1 in-region' note, and added a short note that the runtime-backed `AnthropicBedrock` SDK client is the Messages-API path on `bedrock-runtime`.

### What was verified
- Enforced account rejects `none`/`default`, accepts `aws_review` and `provider_data_share`.
- Mantle (`AnthropicBedrockMantle`) returns 404 for Fable 5.1 in commercial Regions across multiple days.
- All three runtime paths (`AnthropicBedrock` SDK Messages, `invoke_model`, `converse`) work with the `us.`/`global.` CRIS profiles.

Notebook remains valid JSON and the self-verifying capability cells are unchanged.

### DIFF
--- a/notebooks/claude_fable_5_1_getting_started/claude-fable-5-1-getting-started.ipynb
+++ b/notebooks/claude_fable_5_1_getting_started/claude-fable-5-1-getting-started.ipynb
@@ -14,8 +14,8 @@
     "\n",
     "## What you'll learn\n",
     "\n",
-    "- Access setup: the `aws_review` data retention mode this Covered Model requires\n",
-    "- Invoking the model three ways: InvokeModel, Converse, and the Anthropic Messages API on Mantle\n",
+    "- Access setup: the data retention mode this Covered Model requires (`aws_review` or `provider_data_share`)\n",
+    "- Invoking the model: InvokeModel and Converse on the `bedrock-runtime` endpoint\n",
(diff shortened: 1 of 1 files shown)


DanyStinson (commented):
Updates to notebook.

PR #296, opened 2026-09-03
docs: state the dependency and Dependabot policy for samples

scouturier:
Samples are point-in-time examples and are not kept current after
publication. Say so in the README, tell users to refresh dependencies
before deploying, and document in CONTRIBUTING how Dependabot PRs and
alerts are handled.


scouturier:
Closing: this documented a bulk-dismissal policy for Dependabot alerts, which is not the intended approach. Alerts will be addressed rather than dismissed.

